PR #18, opened 2023-05-23
Add request package (CommonMediaRequest and CommonMediaResponse)

bbert:
## Description
Add common API for player requests and responses, following discussion on dash.js: https://github.com/Dash-Industry-Forum/dash.js/issues/4152
This API provides also a plugin request and response callback prototype that can be used as input in every player in order to enable applications to manipulate requests and responses.

A sample test application based on dash.js and using this API will be provided withing dash.js repo.

## Requirements Checklist
- [ ] Unit Tests updated or fixed (N/A)
- [ x] Docs/guides updated


lib/src/request/RequestPlugin.ts
```diff
@@ -0,0 +1,10 @@
+import { CommonMediaRequest } from './CommonMediaRequest.js';
+
+/**
+ * Request plugin API.
+ * @param request - The request to be executed.
+ * @returns A promise that is resolved when the plugin has completed the manipulation of the request.
+ * 
+ * @group Request
+ */
+export type RequestPlugin = (request: CommonMediaRequest) => Promise<any>
```
littlespex, line 10: Design question: Should this function return a new/modified request, or should it modify the incoming request and return `void`?

lib/src/request/CommonMediaResponse.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * The response data.
+   */
+  data?: any
+
+  /**
+   * The network timing of the request/response.
+   */
+  resourceTiming: ResourceTiming
```
littlespex, line 54: Do we need to define our own interface, or can we use `PerformanceResourceTiming` directly?

lib/src/request.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Common Media Request
+ * 
+ * @packageDocumentation
+ */
+export type { ResourceTiming } from './request/ResourceTiming.js';
+export type { CommonMediaRequest } from './request/CommonMediaRequest.js';
+export type { CommonMediaResponse } from './request/CommonMediaResponse.js';
+export type { RequestPlugin } from './request/RequestPlugin.js';
+export type { ResponsePlugin } from './request/ResponsePlugin.js';
```
littlespex, line 10: Is "plugin" the right terminology here? Would `RequestModifier` or `RequestTransform` be more appropriate?

lib/src/request/RequestPlugin.ts
bbert: I would say it is more suitable and flexible to return the new or modified request.

`export type RequestPlugin = (request: CommonMediaRequest) => Promise<CommonMediaRequest>`

lib/src/request/CommonMediaResponse.ts
bbert: The point is that you can't create a `PerformanceResourceTiming` object instance, and the question is: does all browsers support this `PerformanceResourceTiming` interface?
If yes, we can force the players to rely indeed on this interface to retrieve the request metrics instead.

lib/src/request.ts
bbert: I agree, "plugin" is not the most suitable terminology.
What about `RequestInterceptor` and `ResponseInterceptor`? (as in Android OkHttp client)

lib/src/request/CommonMediaResponse.ts
littlespex: Chrome since 29, but Safari came much later at version 11. I think we can use the `ResourceTiming` interface in the PR for now.

lib/src/request/RequestPlugin.ts
littlespex: Agreed.

lib/src/request.ts
littlespex: I'm not familiar with that client, but interceptor is the name used on chromecast to describe similar functionality:

https://developers.google.com/cast/docs/reference/web_receiver/cast.framework.PlayerManager#setMessageInterceptor

This brings up the question: How should the `null` use case be handled? (an interceptor returning a nullish value)

bbert: I've renamed.
`null` use case could be used for enabling the application to force not sending the request, but I don't see the point of doing so.
Should we consider this possiblity ?
As it is (`(request: CommonMediaRequest) => Promise<CommonMediaRequest>`), the interceptor has to return a Promise resolved with a `CommongMediaRequest` object

lib/src/index.ts
```diff
@@ -5,4 +5,5 @@
  */
 export * from './cmcd.js';
 export * from './id3.js';
+export * from './request.js';
```
littlespex, line 8: This line breaks the tests due to a quirk of the build system. The `request` package only exports types/interfaces, and as a result when the Typescript is compiled, there is no `request.js` file. To get around this we need to add the `type` modifier:

```suggestion
export type * from './request.js';
```

littlespex (changes requested):
Everything looks good but there are a few book keeping items to address:
- There needs to be a change log entry for this work. I've edited the PR to point to the 0.5.0 release branch, we can add this work to that release.
- DCO is now enabled for all commits. Follow the instructs here to update the commits for this PR. https://github.com/streaming-video-technology-alliance/common-media-library/pull/18/checks?check_run_id=18033214696
- The tests are not working due to a quirk in the build system. See code suggestion for fix.

dsparacio (approved):

lib/src/index.ts
bbert: Fixed

littlespex (approved):

Issue #103, opened 2024-10-23
Replace Typescript enums with constants


littlespex:
Typescript enums produce non-optimal code when compiled:

```ts
enum CmcdStreamingFormat {
    DASH = 'd',
    HLS = 'h',
    SMOOTH = 's',
    OTHER = 'o',
}
```
```js
var CmcdStreamingFormat;
(function (CmcdStreamingFormat) {
    CmcdStreamingFormat["DASH"] = "d";
    CmcdStreamingFormat["HLS"] = "h";
    CmcdStreamingFormat["SMOOTH"] = "s";
    CmcdStreamingFormat["OTHER"] = "o";
})(CmcdStreamingFormat || (CmcdStreamingFormat = {}));
```

There is another way to express constrained sets using `as const` which produces more concise output (though it's more verbose to author):

```ts
const CmcdStreamingFormat = {
	DASH: 'd',
	HLS: 'h',
	SMOOTH: 's',
	OTHER: 'o',
} as const;
```
```js
const CmcdStreamingFormat = {
    DASH: 'd',
    HLS: 'h',
    SMOOTH: 's',
    OTHER: 'o',
};
```
It would help reduce the code size of production apps using CML if we used the const style. Also of note, this would not constitute a breaking change, and would be transparent to users of CML. The only downside would be that APIs docs might not be as clean.

littlespex:
@cjpillsbury Anything to add on this topic? Do you have any experience documenting the `as const` style?

cjpillsbury:
Here are a few callouts:
1. Discussion from TS folks - https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums (some pitfalls for `Enums` usage are also discussed on this page)
2. Using type structures + data structures can still handle enum benefits in your type relationships pretty easily with a few handy conventions. Below are some examples:

**Derive and exporting the const "lookup object"/"map"'s type**
```ts
export const CmcdStreamingFormat = {
    DASH: 'd',
    HLS: 'h',
    SMOOTH: 's',
    OTHER: 'o',
};

export type CmcdStreamingFormat = typeof CmcdStreamingFormat;
```

**Get the value types to use as a type elsewhere**
```ts
type ValueOf<T> = T[keyof T];
// Or whatever naming convention you prefer here
type CmcdStreamingFormats = ValueOf<CmcdStreamingFormat>;

const someFn = (streamingFormat: CmcdStreamingFormats) => {
  // do stuff;
};
```

PR #108, opened 2024-10-26
fix: refactor enums to const

littlespex:
see: [#103](https://github.com/streaming-video-technology-alliance/common-media-library/issues/103)


lib/src/cmcd/CmcdEncoding.ts
```diff
@@ -1,23 +1,35 @@
...
+ *
  * @beta
  */
-export enum CmcdEncoding {
+export const CmcdEncoding = {
 	/**
 	 * JSON
 	 */
-	JSON = 'json',
+	JSON: CMCD_JSON as typeof CMCD_JSON,
```
littlespex, line 18: This `as` typing will be needed when `isolateDeclarations` is enabled.

lib/src/cmcd/CmcdEncoding.ts
```diff
@@ -1,23 +1,35 @@
+import { CMCD_HEADERS } from './CMCD_HEADERS';
+import { CMCD_JSON } from './CMCD_JSON';
+import { CMCD_QUERY } from './CMCD_QUERY';
+
 /**
  * CMCD encoding types.
  *
  * @group CMCD
  *
+ * @enum
```
littlespex, line 10: Flag these as `enum` so that the API docs only show one API member.

tsconfig.json
```diff
@@ -9,6 +9,7 @@
 		"importHelpers": true,
 		"incremental": true,
 		"inlineSources": true,
+		// "isolatedDeclarations": true,
```
littlespex, line 12: We will enable `isolatedDeclarations` as soon as https://github.com/microsoft/rushstack/issues/4805 is resolved.

lib/src/cmcd/CMCD_HEADERS.ts
```diff
@@ -0,0 +1 @@
+export const CMCD_HEADERS = 'headers' as const;
```
Beraliv, line 1: Given you have a string literal in const variables, you don't need `as const`, you will get string literal type for free

Example on the playground - https://tsplay.dev/WyZZZw

Unfortunately, it doesn't work for let variables, so if you'd like to achieve a consistency in the codebase, this is a good approach then.

Have you added `as const` for consistency or are there other reasons that I'm missing?

lib/src/cmsd/CmsdHeaderField.ts
```diff
@@ -1,24 +1,30 @@
...
 	 */
-	DYNAMIC = 'CMSD-Dynamic',
-}
+	DYNAMIC: CMSD_DYNAMIC as typeof CMSD_DYNAMIC,
+} as const;
+
+/**
+ * @internal
+ */
+export type CmsdHeaderField = typeof CmsdHeaderField[keyof typeof CmsdHeaderField];
```
Beraliv, line 30: As @cjpillsbury suggested in his comment (https://github.com/streaming-video-technology-alliance/common-media-library/issues/103#issuecomment-2436051101), would it be possible to use `ValueOf` in place to avoid duplicated logic in every file?

lib/src/cmcd/CMCD_HEADERS.ts
Beraliv: @littlespex right, given `isolatedDeclaration ` is enabled in tsconfig, when you don't have `as const` on the const variable, would it be resolved as a `string`?

lib/src/cmsd/CmsdHeaderField.ts
littlespex: I wouldn't mind using a helper, but the `ValueOf` helper mentioned in that comment doesn't compile because of circular references, as well as a few others I've tried:
 
Screenshot 2024-10-28 at 09 57 36

If anyone has a solution for this I'd be open to adding it.

cjpillsbury: @littlespex so it's an issue of type vs. value. You can do what you're trying, but I'd lean towards naming the type alias you're constructing differently. Defer to y'all on overall preference. Here's a simple writeup of two different options:

```ts
export const CmcdStreamingFormat = {
    DASH: 'd',
    HLS: 'h',
    SMOOTH: 's',
    OTHER: 'o',
};

// This is the type of the *complex object,* CmcdStreamingFormat
export type CmcdStreamingFormat = typeof CmcdStreamingFormat;

// This should probably be defined + exported in a shared/global types file for the project
export type ValueOf<T> = T[keyof T];

// This is the type of the *values* defined on the complex object, CmcdStreamingFormat
// You *could* do this, but it could get confusing because now your named type definition and your corresponding value declaration have "different shapes"
// Note the "typeof" here
// export type CmcdStreamingFormat = ValueOf<typeof CmcdStreamingFormat>;

// Or you could name the type declaration/alias for the values differently, e.g. making it plural, which allows you to have both types
// and have a corresponding type for the exported lookup/map object, CmcdStreamingFormat
export type CmcdStreamingFormats = ValueOf<CmcdStreamingFormat>;
```

Here's a TS Playground of ☝️:
https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwgOwM7wMIFsEBMDKMUwAhlgJbIDmAYtFifALxwDeAUHF3ACICC+ABIAuOAHJcYgDSduggDL5RYgBbTZXfAFkA8joAqI8anXc4BwQFEASsojqAvgG527APRu4+leVRxfcDAqwIEAnmAhEABmgcFwAFRIWGAANqBwEABGAFbACDBS8XDYeITEZJS09IzsoJCwYRHFOAREpBTUdFAMzI3A0c2lbRWd1TAu7p7eAagqEACuKbhwYFBZJJkpoXCZIbjAUZTAywDUcHXQMMf+yHAkcLMkxLhuVCnrKX1+h2lwUdCxEKrCC5fK1cCXPpwABqJBS82AOiiAB59AA+OAsfQAbQA1sBQgN9ABdCYeLw+PwBIIhGDhSIxGkJABucIRqCK+0OyGuKEBiAgyTSIAyOTyBUGrXKHSq3Rq5IAmgsEkhFrhORBYr4pDt5vByPBVUs4FRgIaUFF5qhKjs8iQrSFkBAAO5wQnzKBwZBka50ppcygG8h8kjIZbuz1IKDEVCQMM21nwvZ5FJPRjB24qEjMkIAIlw5CiUWAxGQ8EeEVQucmcAAchArvzc37+lFc3BgsQaxcGi3JWV2pUuj1MTC2YiUS2BiUpYPRnKYGiyZ4dJ73QK1V6ffy+-sEKmoOmUG44eQSN8AUzE+y4AWiyXgGWtjrgAA6KivuAMXE2g0reGHikOrOj4CAqHcKTvM6fjrjAmpZjmOwNuBLaoDWobLAhIT3FGMZxgW1BQv8npMj2VzLO8EC4vMYBuAwYCiqCEozgOIyyj04L1PAfYscMMrDowfgsLCSZIsivHSkOYxLkAA

tsconfig.json
littlespex: Comments are valid in TS config files, but I've a `TODO` with the a link to the external issue.

lib/src/cmsd/CmsdHeaderField.ts
littlespex: Having different names for the type vs value is something we want to avoid.

littlespex: It would:
- Create a breaking semver change
- Force us to document the `'s'` types separetly
- Having to create in interim type just to use with the helper type seems to defeat the purpose of encapsulating the helper in the first place. 

As mentioned in the inciting issue, the "`keyof typeof`" approach is a bit verbose on the authoring side of things, but it makes things cleaner on the user side.

cjpillsbury: Cool, was simply responding to 
> I wouldn't mind using a helper, but the ValueOf helper mentioned in that comment doesn't compile because of circular references, as well as a few others I've tried

lib/src/cmcd/CMCD_HEADERS.ts
littlespex: At some point during the refactor adding `as const` to exports fixed compile errors I was seeing, but it looks like they are no longer needed. They have been removed from the exports. Thanks for the feedback.

lib/src/cmsd/CmsdHeaderField.ts
littlespex: `ValueOf<typeof CmcdStreamingFormat>` is the answer. Updating now

cjpillsbury:
Same on both counts.

lib/src/cmcd/CmcdEncoding.ts
Beraliv: I'm not familiar `isolatedDeclarations` yet, but can you double-check what error will occur when you remove `as typeof CMCD_JSON`?

I have a feeling that you don't need `as typeof CMCD_JSON` here, because `CMCD_JSON` will already have the right type. However, I can be totally wrong

Beraliv (approved):
Last question about `CMCD_JSON as typeof CMCD_JSON`, otherwise LGTM!

lib/src/cmcd/CmcdEncoding.ts
littlespex: Screenshot 2024-10-29 at 07 58 28

littlespex: If you have any other questions about this feature check out [the release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations).

Beraliv: Thanks, feel free to resolve this conversation!

Issue #119, opened 2024-10-29
Remove CMAF HAM dependencies


littlespex:
The dependencies imported by the CMAF HAM library are not used anywhere else in the library, but since they are listed as `dependencies`, they are install for anyone using CML. In the future we will deliver discreet packages for each CML feature (CMCD, 608, HAM, etc), but for now, update the HAM utils to include it's dependencies via injection.

PR #120, opened 2024-10-29
fix: move CMAF HAM dependencies from lib to samples

littlespex:
see: #119


Beraliv (dismissed):
Clear change, no questions so far

nicolaslevy:
Agree! with @Beraliv. I like the new way to use external parsers

PR #126, opened 2025-02-05
feat: add xml parsing utility function 

### LINKED ISSUE #125: Add XML parsing utils
In order to build features like DASH manifest and IMSC caption parsers, CML needs a XML parsing utility. Looking at the various options out there, [tXML](https://github.com/TobiasNickel/tXml) seems like a good choice that is performant on older/resource constrained devices. Unfortunately it looks like this project is no longer being actively maintained (https://github.com/TobiasNickel/tXml/issues/53). After discussing it with the group we have determined that we would like to maintain our own version internally, while incorporating some of the changes made in [`dash.js`'s](https://github.com/Dash-Industry-Forum/dash.js/blob/development/externals/tXml.js) and [`shaka-player`'s](https://github.com/shaka-project/shaka-player/blob/main/lib/util/tXml.js) internal versions of the library.

littlespex:
## Description
This is a port of the `tXml` XML parsing utility. For now it was ported as is, but in the future we could look into:
- Parse namespaces
- Document level helper functions.
- Review `dash.js` and `shaka-player` implementation for common functionality that was added in their clones of the library.

Resolves #125. 

## Requirements Checklist
- [x] Docs/guides updated


dsparacio (dismissed):

lib/src/xml/parseXml.ts
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * @example
+ * ```ts
+ * import { parseXml } from '@svta/common-media-library/xml/parseXml.js';
+ *
+ * const obj = parseXml('<root><child>text</child></root>');
+ * console.log(obj[0].children[0].tagName);
+ * // -> "child"
+ * console.log(obj[0].children[0].children[0]);
+ * // -> "text"
```
littlespex, line 24:
```suggestion
 * const obj = parseXml('<root><child>text</child></root>');
 * console.log(obj[0].childNodes[0].nodeName);
 * // -> "child"
 * console.log(obj[0].childNodes[0].childNodes[0].nodeValue);
 * // -> "text"
```

Issue #163, opened 2025-03-31
CommonMediaRequest / CommonMediaResponse Updates


littlespex:
There are a few changes needed in the `CommonMediaRequest` / `CommonMediaResponse` types:
1. The `CommonMediaRequest` types does not define a `body` property. This is needed for `POST` requests which are used in things like DRM license requests. Add this property to the type.
2. The `CommonMediaResponse` currently requires `resourceTiming` to be defined. We have found certain platforms that do not easily provide this information. While it can be manually calculated, it may be better to define this property as optional, like all of the other properties in the type.

PR #164, opened 2025-03-31
feat: update request types

littlespex:
see: #163

### DIFF
--- a/lib/src/request/CommonMediaRequest.ts
+++ b/lib/src/request/CommonMediaRequest.ts
@@ -1,4 +1,5 @@
 import type { Cmcd } from '../cmcd/Cmcd.js';
+import type { RequestType } from './RequestType.js';
 
 /**
  * Common request API.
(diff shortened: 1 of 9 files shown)


lib/src/request/CommonMediaRequest.ts
```diff
@@ -12,46 +13,51 @@ export type CommonMediaRequest = {
...
 	/**
 	 * The request's method (GET, POST, etc).
 	 */
-	method: string
+	method?: string;
+
+	/**
+	 * The body of the request.
+	 */
+	body?: BodyInit;
```
littlespex, line 26: @agajassi Will this work for your DRM changes?

lib/src/request/CommonMediaRequest.ts
```diff
@@ -12,46 +13,51 @@ export type CommonMediaRequest = {
...
 
 	/**
 	 * The response type with which the response from the server shall be compatible.
 	 */
-	responseType?: string
+	responseType?: RequestType;
```
littlespex, line 31: Having a more strict type for `responseType` allows us to map the `data` property of `CommonMediaResponse` to a specific data type.

lib/src/request/ResponseTypeMap.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @group Request
+ *
+ * @beta
+ */
+export type ResponseTypeMap<T extends string | undefined> =
+	T extends 'json' ? any :
+	T extends 'text' ? string :
+	T extends 'blob' ? Blob :
+	T extends 'arraybuffer' ? ArrayBuffer :
+	T extends 'document' ? XmlNode :
```
littlespex, line 15: We are using the simpler `XmlNode` type here instead of [`XMLDocument`](https://developer.mozilla.org/en-US/docs/Web/API/XMLDocument) so that CML's `parseXml` function can also be used.

lib/src/request/CommonMediaRequest.ts
agajassi: Yes. Thanks. That is exactly what I needed.

agajassi (approved):

PR #194, opened 2025-05-31
task/restore-missing-tests

littlespex:
## Description
Restores tests that were being skipped due to an incorrectly configured glob filter

## Requirements Checklist
- [x] Docs/guides updated

### DIFF
--- a/lib/test/cmaf/ham/init.ts
+++ b/lib/test/cmaf/ham/init.ts
@@ -1,0 +1,34 @@
+import { setDashParser, setDashSerializer, setHlsParser, type DashManifest, type HlsManifest } from '@svta/common-media-library';
+import { Parser } from 'm3u8-parser';
+import { Builder, parseString } from 'xml2js';
+
+setDashParser((raw: string) => {
+	let parsed: DashManifest;
+	parseString(raw, (err: Error | null, result: DashManifest) => {
+		if (err) {
+			throw new Error(err.message);
+		}
+		parsed = result as DashManifest;
+	});
+	// @ts-ignore
+	return parsed;
+});
+
+setHlsParser((text: string) => {
+	const parser = new Parser();
+
+	parser.push(text);
+	parser.end();
+	const parsedHlsManifest = parser.manifest as any;
+	if (!parsedHlsManifest) {
+		throw new Error();
+	}
+
+	return parsedHlsManifest as HlsManifest;
+});
+
+
+setDashSerializer((manifest: DashManifest) => {
+	const builder = new Builder();
+	return builder.buildObject(manifest);
+});
--- a/lib/test/cmaf/ham/init.ts
+++ b/lib/test/cmaf/ham/init.ts
@@ -4,7 +4,7 @@ import { Builder, parseString } from 'xml2js';
 
 setDashParser((raw: string) => {
 	// `parsed` is populated by the synchronous `parseString` callback
-	// so we need to explicitlytell Typescript that it is not undefined.
+	// so we need to explicitly tell Typescript that it is not undefined.
 	let parsed!: DashManifest;
 
 	parseString(raw, (err: Error | null, result: DashManifest) => {
--- a/lib/test/cmaf/ham/init.ts
+++ b/lib/test/cmaf/ham/init.ts
@@ -3,14 +3,18 @@ import { Parser } from 'm3u8-parser';
 import { Builder, parseString } from 'xml2js';
 
 setDashParser((raw: string) => {
-	let parsed: DashManifest;
+	// `parsed` is populated by the synchronous `parseString` callback
+	// so we need to explicitlytell Typescript that it is not undefined.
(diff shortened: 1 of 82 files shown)


lib/test/cmaf/ham/init.ts
```diff
@@ -0,0 +1,34 @@
...
+
+setDashParser((raw: string) => {
+	let parsed: DashManifest;
+	parseString(raw, (err: Error | null, result: DashManifest) => {
+		if (err) {
+			throw new Error(err.message);
+		}
+		parsed = result as DashManifest;
+	});
+	// @ts-ignore
```
Beraliv, line 13: Why do you need ts-ignore? If it's still useful (unmatched types), what about replacing it with ts-expect-error?

```suggestion
	// @ts-expect-error
```

I also encourage adding the reason after the comment, with the why

lib/test/cmaf/ham/init.ts
```diff
@@ -0,0 +1,34 @@
...
+	return parsed;
+});
+
+setHlsParser((text: string) => {
+	const parser = new Parser();
+
+	parser.push(text);
+	parser.end();
+	const parsedHlsManifest = parser.manifest as any;
```
Beraliv, line 22: Is it worth moving the type here, so the return statement will be without type assertion?

```suggestion
	const parsedHlsManifest = parser.manifest as HlsManifest | undefined;
```

lib/test/cmaf/ham/init.ts
```diff
@@ -0,0 +1,34 @@
...
+	if (!parsedHlsManifest) {
+		throw new Error();
+	}
+
+	return parsedHlsManifest as HlsManifest;
```
Beraliv, line 27:
```suggestion
	return parsedHlsManifest;
```

lib/test/drm/cenc/getPSSHData.test.ts
```diff
@@ -1,7 +1,7 @@
-import { getPSSHData } from '@svta/common-media-library/drm/cenc/getPSSHData';
+import { getPSSHData } from '@svta/common-media-library/drm';
 import { strictEqual } from 'node:assert';
 import { describe, it } from 'node:test';
-import { samplePsshBox } from '../common/samplePsshBox';
+import { samplePsshBox } from '../common/samplePsshBox.ts';
```
Beraliv, line 4: I thought that we have to postfix imports with .js, not .ts. Am I missing something?

Beraliv (commented):
Left some comments, mostly around the types

lib/test/cmaf/ham/init.ts
littlespex: The error it's ignoring is: `Variable 'parsed' is used before being assigned.ts(2454)`. `parsed` is populated by the synchronous `parseString` callback.

lib/test/cmaf/ham/init.ts
littlespex: That change won't compile: `Conversion of type Manifest to type HlsManfiest may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to unknown first.`.

littlespex: Added a comment describing the need for the forced casting.

lib/test/drm/cenc/getPSSHData.test.ts
littlespex: The tests are run with Node's type-stripping feature, so in the test files you need to import the files with their proper file extension. File imported from the `test` directory will have `.ts`, and files imported from the compiled library with have `.js`.

lib/test/drm/cenc/getLicenseServerUrlFromContentProtection.test.ts
```diff
@@ -43,13 +40,13 @@ describe('getLicenseServerUrlFromContentProtection', () => {
 	});
 
 	it('should return license server URL for clearkey prefix', () => {
-		contentProtection[0].laUrl.__prefix = 'clearkey';
+		// contentProtection[0].laUrl.__prefix = 'clearkey';
```
littlespex, line 43: @agajassi What is the purpose of the `__prefix` code? I don't see the `getLicenseServerUrlFromContentProtection` using that property in its logic. Can we remove these lines?

samples/cmaf-ham-conversion/src/index.ts
```diff
@@ -27,12 +28,12 @@ setHlsParser((text: string) => {
 
 	parser.push(text);
 	parser.end();
-	const parsedHlsManifest = parser.manifest;
+	const parsedHlsManifest = parser.manifest as any;
```
Beraliv, line 31: (Optional) I suggest avoiding `as any` where possible

littlespex: The only alternative that I know is `as unknown`. Is that any better?

Issue #193, opened 2025-05-28
VTT Parsing Utilities


littlespex:
Add a VTT parser and utilities to CML. There are many forks of the original [mozilla/vtt.js](https://github.com/mozilla/vtt.js), but this seems to be the most actively maintained:

https://github.com/videojs/vtt.js

The main branch has the most recent code, but here are additional changes in a few branches that may be helpful:

- VideoJS is using this branch: https://github.com/videojs/vtt.js/tree/0.15.x
- Adds regions: https://github.com/videojs/vtt.js/pull/50
- WIP style block support: https://github.com/videojs/vtt.js/pull/35
- Style overrides: https://github.com/videojs/vtt.js/pull/27

PR #196, opened 2025-06-03
feat: add support for web vtt parsing

littlespex:
Resolves: #193

> [!NOTE]
> This is only the parsing functionality. Rendering and processing of VTTCues will be addressed later.

This PR pulls heavily from [videojs/vtt.js](https://github.com/videojs/vtt.js). A few notable changes that should be considered breaking changes when migrating from `vtt.js`:
- The parser has been updated to support the latest version of the [WebVTT spec](https://www.w3.org/TR/webvtt1/)
  - Parses `REGION` blocks
  - Parses `STYLE` blocks. Styles are parsed as strings. Full CSS parsing should be handled at the application layer for now. 
  - Uses the latest values and defaults for VTTCue settings like `position`, `positionAlign`, `align`, etc.
-  The `decoder` option has been removed from the `WebVttParser` constructor arguments. Data that needs be decoded should be handled before being passed to `parse()` (i.e. `TextDecoder` or `TextDecoderStream`).
- Polyfills for `VTTCue` and `VTTRegion` are not provided. The parser allows for factory functions to be passed in via the options object for cases where `VTTCue` and `VTTRegion` don't exist globally, i.e. Node or Web Workers.

New features:
- Simple `parseWebVtt` function.
- A transform stream `WebVttTransformStream`

## Requirements Checklist
- [x] Docs/guides updated


agajassi (dismissed):

dsparacio (dismissed):

lib/src/webvtt/WebVttCue.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export type WebVttCue = {
+	id: string;
+	startTime: number;
+	endTime: number;
+	pauseOnExit: boolean;
+	text: string;
+	align: AlignSetting;
+	region: WebVttRegion | null;
+	vertical: DirectionSetting;
```
Beraliv, line 18: This type is defined in TypeScript as `type DirectionSetting = "" | "lr" | "rl";`. Is it expected to use this type, or something else meant to be imported?

lib/src/webvtt/WebVttParser.ts
```diff
@@ -0,0 +1,459 @@
...
+	private regionSettings: Settings | null = null;
+	private style: string;
+	private cue: WebVttCue | null = null;
+	private createCue: WebVttCueFactory;
+	private createRegion: WebVttRegionFactory;
+
+	/**
+	 * A callback function that is called when a parsing error occurs.
+	 */
+	onparsingerror: ((error: WebVttParsingError) => void) | undefined;
```
Beraliv, line 38: Is there a reason not to use `?:` syntax, for example:

```suggestion
	onparsingerror?: (error: WebVttParsingError) => void;
```

Similar question to other `on<handler>` functions

lib/src/webvtt/parse/parseCue.ts
```diff
@@ -0,0 +1,124 @@
...
+import { Settings } from './Settings.js';
+
+const BAD_TIMESTAMP = 'Malformed time stamp.';
+
+export function parseCue(input: string, cue: WebVttCue, regionList: WebVttRegion[]): void {
+	// Remember the original input if we need to throw an error.
+	var oInput = input;
+	// 4.1 WebVTT timestamp
+	function consumeTimeStamp() {
+		var ts = parseTimeStamp(input);
```
Beraliv, line 15: (Optional) I've seen vars, should we use `const` instead?

```suggestion
	const oInput = input;
	// 4.1 WebVTT timestamp
	function consumeTimeStamp() {
		const ts = parseTimeStamp(input);
```

Similar below, I'll leave this with you

lib/src/webvtt/WebVttCue.ts
littlespex: It is expected to use that type. It is from the `DOM` ts library and is a reference to the [`vertical`](https://developer.mozilla.org/en-US/docs/Web/API/VTTCue/vertical) properties values.

Beraliv (approved):
Amazing job, @littlespex! I left some comments, but I don't want to block your progress, hence, approving in advance

lib/src/webvtt/WebVttCue.ts
Beraliv: Right, the name didn't look aligned, so if it's expected, no concerns anymore

lib/src/webvtt/parse/parseCue.ts
littlespex: Fixed.

lib/src/webvtt/WebVttParser.ts
littlespex: Fixed.

lib/src/webvtt/WebVttParser.ts
```diff
@@ -0,0 +1,466 @@
...
+				// This line contains HLS X-TIMESTAMP-MAP metadata
+				parseOptions(input, (k: string, v: string): void => {
+					switch (k) {
+						case 'X-TIMESTAMP-MAP':
+							parseTimestampMap(v);
+							break;
+					}
+				}, /=/);
+			}
+		};
```
littlespex, line 154: @gkatsev `X-TIMESTAMP-MAP` is mentioned in the HLS spec, but not the latest WebVTT spec. I don't see anything in the WebVTT spec that would account for this line, except maybe the mention of ["file magic" in section 4.1 (item #4)] (https://www.w3.org/TR/webvtt1/#file-structure). Do you know if the concept of ["metadata headers"](https://www.w3.org/TR/2015/WD-webvtt1-20151208/#webvtt-metadata-header) still exists?

gkatsev: it's worth keeping around for hls support. In WebVTT, we want to bring back those headers as an extension point for HLS and others https://github.com/w3c/webvtt/issues/485#issuecomment-646089841, but it hasn't happened yet. Once it comes back to WebVTT the HLS spec will update to point it but would continue defining it as WebVTT would just say that there can be data in that location and not what it says.

dloftus (approved):

gkatsev:
Late to the party, but not seeing any issues, and took `parseWebVtt` method for a quick spin in node 24 and LGTM!

littlespex:
@gkatsev Thanks for taking a look. We are preparing a release and wanted to get this in. To your slack suggestion about the parsing function possibly being `async`: The `WebVttParser` is still a sax-style parser with events for every cue/region/style encountered, so long files can be chunked. We've also added a `WebVttTransformStream` for modern environments that support the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API).

Issue #202, opened 2025-06-08
The ISO BMFF utilities are not properly typed


littlespex:
The ISO BMFF utilities are all typed with `Box<any>`, which is fundamentally the same as just typing them as `any`. This results in the end user needing to force cast all return values.

PR #203, opened 2025-06-08
fix: iso bmff utilities are not properly typed

littlespex:
Resolves: #202

This PR fixes a number of issues in the ISO BMFF library:
- Proper types for all boxes are provided in `@svta/common-media-library/iso/bmff/boxes`
- All utilities and parsers now use proper box types. 
- Fixed inconsistent property and type naming
- Refactored all of the utility functions to use the same argument order, with the data element as the first argument, and the parser config as the last.

## Requirements Checklist
- [x] Docs/guides updated


lib/src/iso/bmff/boxes/AudioRenderingIndicationBox.ts
```diff
@@ -0,0 +1,13 @@
+import type { FullBox } from './FullBox.js';
+
+/**
+ * ISO/IEC 14496-12:202x - 12.2.8 Audio rendering indication box
+ *
+ * @group ISOBMFF
+ *
+ * @beta
+ */
+export type AudioRenderingIndicationBox = FullBox & {
```
Beraliv, line 10: (optional) potentially useful to prettify the intersection by using `Prettify`, where `Prettify` is:

```ts
export type Prettify<Type> = Type extends Function
  ? Type
  : Extract<
      {
        [Key in keyof Type]: Type[Key];
      },
      Type
    >;
```

See examples in https://github.com/ts-essentials/ts-essentials/blob/master/lib/prettify/README.md

lib/src/iso/bmff/boxes/VisualSampleEntryBox.ts
```diff
@@ -0,0 +1,25 @@
+import type { SampleEntryBox } from './SampleEntryBox.js';
+
+/**
+ * ISO/IEC 14496-15:2014 - 12.1.3.1 avc1/2/3/4, hev1, hvc1, encv
+ *
+ * @group ISOBMFF
+ *
+ * @beta
+ */
+export type VisualSampleEntryBox<T = 'avc1' | 'hev1' | 'hvc1' | 'encv'> = SampleEntryBox & {
```
Beraliv, line 10: Is it meant to be with `extends` and you have default type

```suggestion
export type VisualSampleEntryBox<T extends 'avc1' | 'hev1' | 'hvc1' | 'encv' = <default_type>> = SampleEntryBox & {
```

or multiple types are intentional?

Beraliv (dismissed):
I left some optional comments, but I don't have major concerns. Great work, approving in advance!

Beraliv (approved):
Great work!

PR #213, opened 2025-06-13
ci(lint): func-style - declaration function

Beraliv:
## Description
Please describe the changes:
- If it's a feature or enhancement please be as detailed as possible.
- If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.
- If this change includes work derived from another project, please:
    - Include links to the original project and the specific files that were used.
		- Provide the appropriate attribution in the `NOTICE` file, including the full license text if available.
		- Update "Thanks" section of the `README.md`.

## Requirements Checklist
- [ ] All commits have DCO sign-off from the author
- [ ] Unit Tests updated or fixed
- [ ] Docs/guides updated


Beraliv:
> My original comment was directed at named exports. I don't think we need to disallow arrow functions completely. Can we update the rule to only apply to exports?

Ah, okay, I'll check it out

lib/src/cta/608/findCta608Nalus.ts
```diff
@@ -14,7 +14,7 @@ export function findCta608Nalus(raw: DataView, startPos: number, size: number):
 	const cta608NaluRanges = [];
 
 	// Check SEI data according to ANSI-SCTE 128
-	const isCTA608SEI = function (payloadType: number, payloadSize: number, raw: DataView, pos: number) {
+	const isCTA608SEI = (payloadType: number, payloadSize: number, raw: DataView, pos: number) => {
```
Beraliv, line 17: avoided mixed arrow function / declaration functions, and simply using arrow functions here

littlespex (approved):

PR #234, opened 2025-07-18
feat: add UUID utilities for converting to and from hex strings and changing endianness

littlespex:
## Requirements Checklist
- [x] Docs/guides updated

### DIFF
--- a/lib/config/common-media-library.api.md
+++ b/lib/config/common-media-library.api.md
@@ -60,6 +60,9 @@ export class ArithmeticMeanEstimator implements ThroughputEstimator {
 // @beta
 export function arrayBufferToString(arrayBuffer: ArrayBuffer, encoding: Encoding): string;
 
+// @beta
+export function arrayBufferToUuid(buffer: ArrayBuffer): string;
+
 // @alpha
 export type AudioChannelConfiguration = {
     $: {
@@ -833,6 +836,9 @@ export type DecodedId3Frame<T> = {
 export function decodeFairPlayLicense(response: string | ArrayBuffer): Uint8Array;
 
 // @beta
+export function decodeHex(hex: string): ArrayBuffer;
+
+// @beta
 export function decodeIso8601Duration(isoDuration: string): number;
 
 // @beta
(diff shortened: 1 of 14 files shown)


lib/config/common-media-library.api.md
```diff
@@ -832,6 +835,9 @@ export type DecodedId3Frame<T> = {
 // @beta
 export function decodeFairPlayLicense(response: string | ArrayBuffer): Uint8Array;
 
+// @beta
+export function decodeHex(hex: string): ArrayBuffer;
```
Beraliv, line 839: Is it worth adding 
```suggestion
export function decodeHexToArrayBuffer(hex: string): ArrayBuffer;
```

It will be useful if you will have something else (e.g. Int8Array) in the future and you won't need to rename it and deprecation.

Same below for other functions.

What do you think?

lib/test/drm/playready/toBigEndianKeyId.test.ts
```diff
@@ -0,0 +1,42 @@
...
+		equal(result, longBuffer);
+	});
+
+	it('should only swap the first 8 bytes (GUID portion)', () => {
+		const input = uuidToArrayBuffer('00000000-0000-0000-1234-567890abcdef');
+		const result = toBigEndianKeyId(input);
+		const resultHex = arrayBufferToUuid(result);
+
+		// First 8 bytes should be swapped, last 8 bytes should remain unchanged
+		equal(resultHex, '00000000-0000-0000-1234-567890abcdef');
```
Beraliv, line 40: Sorry, a bit confused, but do we expect anything to change in this test case? Hex string is still the same

Beraliv (dismissed):
The question and the suggestion, otherwise, LGTM

lib/test/drm/playready/toBigEndianKeyId.test.ts
littlespex: This was a suggested test, and its use case is already covered by the other tests. Removing it.

PR #246, opened 2025-08-27
issue/244-consolidate-decode

### LINKED ISSUE #244: Consolidate text decoding utilities
There are a number of text decoding helper utilities, some of which seem to do the same thing, just with different input types (ArrayBuffer, DataView, TypedArray). We should consolidate these into a single function that uses `TextDecoder` by default, with a fallback for platforms that do not have a `TextDecoder` implementation.

littlespex:
## Description

Consolidate various text decoding utilities into a single `decodeText` function.

Resolves #244 

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated


lib/src/utils/decodeText.ts
```diff
@@ -0,0 +1,153 @@
...
+
+				// Single byte (ASCII)
+				if (char < 128) {
+					byteOffset++;
+				}
+				// 2-byte sequence
+				else if (char >= 194 && char <= 223) {
+					if (byteOffset + 1 < byteLength) {
+						const byte2 = view.getUint8(byteOffset + 1);
+						if (byte2 >= 128 && byte2 <= 191) {
```
Beraliv, line 82: This condition repeats 5-6 times later, is it worth creating utility function for it?

lib/test/utils/decoderTest.ts
```diff
@@ -0,0 +1,21 @@
+import { after, before, describe } from 'node:test';
+
+const TD = TextDecoder;
```
Beraliv, line 3: (Optional) more meaningful name, e.g. NativeTextDecoder?

```suggestion
const NativeTextDecoder = TextDecoder;
```

Wdyt?

Beraliv (approved):
Great work, LGTM!

lib/src/utils/decodeText.ts
littlespex: It's slightly faster inline, so it would be a trade off.

Beraliv: Right, makes sense

PR #344, opened 2026-04-12
refactor(c2pa): pre-publish review fixes

littlespex:
## Summary

Pre-publish code review of the `@svta/cml-c2pa` package, addressing performance, security, tree-shaking, and consistency issues across 13 files.

### Performance

- **Eliminate triple BMFF parsing in init validation** — `validateC2paInitSegment` was calling `readIsoBoxes()` three times on the same bytes: once directly for the `mdat` check, once via `readC2paManifest`, and once via `extractManifestCertificate`. Now parses once and passes the pre-parsed boxes through. The certificate is extracted from the already-available signature bytes instead of re-navigating the entire JUMBF structure. (`validateC2paInitSegment.ts`, `readC2paManifest.ts`, `extractManifestCertificate.ts`)

- **Replace `bytesToHex` with lookup table** — The previous implementation created an intermediate `Array`, mapped each byte to a hex string, then joined. The new version uses a pre-built 256-entry lookup table and string concatenation, avoiding per-call allocations. (`utils.ts`)

- **Hoist `TextEncoder`/`TextDecoder` to module scope** — Five call sites were creating new encoder/decoder instances on every invocation. These are now module-level constants. In `buildSigStructure.ts`, the `"Signature1"` context string is encoded once at module load rather than on every signature operation. (`buildSigStructure.ts`, `parseJumbfLabel.ts`, `extractCertificateInfo.ts`, `readC2paManifest.ts`)

### Security

- **Constant-time hash comparison for assertion hashes** — `validateAssertionHashes.ts` was using early-exit byte comparison, while `validateBmffHash.ts` correctly used constant-time comparison. Extracted the constant-time `hashesEqual` into `utils.ts` and replaced both call sites, closing a potential timing side-channel in assertion hash verification. (`utils.ts`, `validateBmffHash.ts`, `validateAssertionHashes.ts`)

### Tree-shaking

- **Use `cbor-x` subpath imports** — `validateC2paInitSegment.ts` imported from bare `cbor-x` (pulling both encoder and decoder), while all other files used the tree-shakeable `cbor-x/decode` and `cbor-x/encode` subpaths. Aligned to use subpath imports consistently. (`validateC2paInitSegment.ts`)

### Correctness

- **Add RSA-PSS support to `resolveImportAlgorithm`** — The function only handled OKP and EC key types, falling through to ECDSA for anything else. If a session key ever uses RSA-PSS (which the COSE algorithm map already supports), the import would fail with an incorrect algorithm. Now correctly maps `kty: 'RSA'` to `RSA-PSS` with the appropriate hash. (`resolveImportAlgorithm.ts`)

- **Fix stale TSDoc window size** — The `validateSequenceNumber` TSDoc claimed a sliding window of "1 000" sequence numbers, but the actual constant is `SEEN_SEQUENCES_WINDOW_SIZE = 32`. (`validateSequenceNumber.ts`)

### Package metadata

- **Update package description** — Changed from "manifest parsing" to "live video validation for BMFF/MP4 containers" to accurately reflect the package's scope (cryptographic validation, session key verification, sequence tracking). (`package.json`)

- **Rename `NOTICE` to `NOTICE.md`** — Aligns with the naming convention used by all other packages in the monorepo (`webvtt`, `iso-bmff`, etc.). Updated `files` array in `package.json`. (`NOTICE.md`, `package.json`)

## Test plan

- [ ] `npm run build -w libs/c2pa` — builds cleanly, API Extractor passes
- [ ] `npm test -w libs/c2pa` — all 121 tests pass
- [ ] `npm run typecheck` — clean
- [ ] `npm run format` — no lint errors


littlespex:
@N1Knight

libs/c2pa/src/cose/resolveImportAlgorithm.ts
```diff
@@ -1,7 +1,13 @@
...
 
-export function resolveImportAlgorithm(jwk: { kty: string; crv: string }): AlgorithmIdentifier | EcKeyImportParams {
-	return jwk.kty === KEY_TYPE_OKP ? { name: jwk.crv } : { name: ECDSA_ALGORITHM, namedCurve: jwk.crv }
+export function resolveImportAlgorithm(jwk: { kty: string; crv: string; alg?: string }): AlgorithmIdentifier | EcKeyImportParams | RsaHashedImportParams {
+	if (jwk.kty === KEY_TYPE_OKP) return { name: jwk.crv }
+	if (jwk.kty === KEY_TYPE_RSA) {
+		const hash = jwk.alg === 'PS384' ? 'SHA-384' : jwk.alg === 'PS512' ? 'SHA-512' : 'SHA-256'
+		return { name: RSA_PSS_ALGORITHM, hash: { name: hash } }
+	}
+	return { name: ECDSA_ALGORITHM, namedCurve: jwk.crv }
```
N1Knight: The spec (§18.25.2) does technically allow any COSE_Key here, but the canonical example uses EC2/P-256 and the verifiable-segment-info-map travels in every segment, so RSA signatures (~256B vs ~64B for EC) are a bad fit size-wise for this method. In practice this code path is currently only reachable for EC/OKP: convertCoseKeyToJwk only emits those, and CoseKeyJwk doesn't carry alg, so the PS384/PS512 check here would always fall through to SHA-256.
I think it's cleaner to revert to the EC/OKP-only version for now and track RSA-PSS support as a separate issue — it'd also need changes in CoseKeyJwk (discriminated union), convertCoseKeyToJwk (RSA kty + alg mapping per RFC 8230), and an end-to-end test vector. Happy to pick that up if there's a real use case asking for it. What do you think?

littlespex: I agree. We can revert this, and if a real world use case for RSA comes up in the future, we can add the feature then.

littlespex: Reverted to the EC/OKP-only version in e239f1aa. RSA-PSS support can be tracked as a separate issue if a real use case comes up.

nicolaslevy (approved):

PR #378, opened 2026-07-17
feat(c2pa): add VOD Merkle segment validation

valentinamgiusti:
## Summary

Adds VOD Merkle validation (C2PA 2.4 §15.12.2.2 / §18.6) to `@svta/cml-c2pa`, released as 1.1.0:

- Implement `validateC2paMerkleSegment(segmentBytes, merkleMaps, state?)`: verifies a fragmented MP4 media segment against the merkle maps stored in the init manifest. Parses the auxiliary uuid box (both the spec §A.5.4 format and the JUMBF format), recomputes the offset-prefixed leaf hash, walks the Merkle proof up to the row stored in the manifest, and enforces `location` continuity via caller-held state.
- `validateC2paInitSegment` now extracts `merkleMaps` from the `c2pa.hash.bmff.v3` assertion and validates each entry's `initHash` binding against the raw init bytes. VOD Merkle streams no longer report `SESSIONKEY_INVALID`.
- New standard status codes: `assertion.bmffHash.malformed` / `assertion.bmffHash.mismatch`.
- Shared helpers extracted first (`parseExclusions`, uuid/byte utils) with no behavior change for the existing VSI and manifest-box paths.

Structured so each commit passes the full test suite on its own (helpers refactor, merkle module, init integration, release chore).

## Test plan

- [x] `npm run build -w=libs/c2pa` compiles and API Extractor passes
- [x] `npm test -w=libs/c2pa`: 146/146 tests pass, including new unit and proof-path suites under `test/merkle/`
- [x] Validated end-to-end against a real VOD stream re-signed with c2patool (all segment hashes match; per-segment proofs verify against the init manifest row)


libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	const layers = treeLayout(count)
+	if (!layers.includes(manifestRow.length)) {
+		return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
+	}
+
+	const treeDepth = layers.length - 1
+	const path = (proofPath ?? []).filter((node): node is Uint8Array => node !== null)
+	if (path.length > treeDepth) return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
```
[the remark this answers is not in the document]
valentinamgiusti: Fixed in 742dc93.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,293 @@
...
+			if (!sibling) return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
+			currentHash = currentIndex % 2 === 0
+				? await hashPair(currentHash, sibling, alg)
+				: await hashPair(sibling, currentHash, alg)
+		}
+		currentIndex = Math.floor(currentIndex / 2)
+	}
+
+	const expected = manifestRow[currentIndex]
+	if (!expected) return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
```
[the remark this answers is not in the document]
valentinamgiusti: Fixed in 742dc93.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,297 @@
...
+type MerkleValidationCode = LiveVideoStatusCode | C2paStatusCode
+
+const MERKLE_BOX_PURPOSE = 'merkle'
+
+// --- Auxiliary uuid box extraction ---
+
+type AuxPayload = Uint8Array | 'other' | 'malformed'
+
+// §A.5.4 format: CBOR map { box_purpose, data }.
+function readSpecAuxPayload(rawPayload: Uint8Array): AuxPayload {
```
littlespex: **Main question on this PR:** this format (bare CBOR map `{ box_purpose, data }`) doesn't match §A.5.4 in either C2PA 2.3 or 2.4. The spec's auxiliary box is the same `ContentProvenanceBox` as the manifest box: `FullBox('uuid', extended_type = d8fec3d6-1b0e-483c-9297-5828877ec481)` with version/flags, a null-terminated `box_purpose` *string*, then raw CBOR (`bmff-merkle-map`) — i.e. exactly what `readJumbfAuxPayload` below parses (and what c2pa-rs writes). So the "JUMBF format" fallback is actually the spec format, and this branch parses a layout I couldn't find in any spec version or implementation (a GitHub-wide code search for the `1b0d-11ec` UUID finds nothing).

If a draft or tool emits this format, could you cite it here? Otherwise I'd remove this function, `MERKLE_AUX_UUID`, and the integer-key path in `readMapField`, and relabel `readJumbfAuxPayload` as the §A.5.4 format. As written, the default test fixture (`buildMerkleAuxBox`) exercises this format rather than the one real content uses.

libs/c2pa/src/utils.ts
```diff
@@ -64,15 +64,38 @@ const C2PA_MANIFEST_UUID: readonly number[] = [
...
 // JUMBF UUID per ISO 19566-5 (used by c2pa-rs and other JUMBF-compliant tools)
-const JUMBF_UUID: readonly number[] = [
+/** @internal */
+export const JUMBF_UUID: readonly number[] = [
 	0xd8, 0xfe, 0xc3, 0xd6, 0x1b, 0x0e, 0x48, 0x3c,
 	0x92, 0x97, 0x58, 0x28, 0x87, 0x7e, 0xc4, 0x81,
 ]
 
-function matchesUuid(usertype: readonly number[], expected: readonly number[]): boolean {
+// C2PA auxiliary uuid box extended type per §A.5.4: d8fec3d6-1b0d-11ec-8100-000000000000
```
littlespex: I couldn't find this UUID in C2PA 2.3 or 2.4. §A.5.1.1 defines the ContentProvenanceBox extended type as `d8fec3d6-1b0e-483c-9297-5828877ec481` (the constant above labeled `JUMBF_UUID`), and §A.5.4's merkle box reuses that same box type. If there's no producer emitting this UUID, suggest removing it (see the comment on `readSpecAuxPayload`).

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,297 @@
...
+
+type BmffMerkleMapSegment = {
+	readonly uniqueId: number
+	readonly localId: number
+	readonly location: number
+	/** Sibling proof path; null when the manifest stores the leaf row */
+	readonly hashes: readonly (Uint8Array | null)[] | null
+}
+
+// §18.6 integer keys; string keys accepted as fallback.
```
littlespex: The `bmff-merkle-map` CDDL uses string keys (`"uniqueId"`, `"localId"`, `"location"`, `"hashes"`) in both 2.3 (§A.5.4.2) and 2.4 — integer keys don't appear in either version, so this comment has the primary/fallback relationship inverted. If no known producer emits integer keys, consider dropping that path; at minimum the comment should cite where they come from.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,297 @@
...
+function readMapField(map: unknown, intKey: number, name: string): unknown {
+	if (map instanceof Map) return map.get(intKey) ?? map.get(name)
+	const record = map as Record<string | number, unknown>
+	return record[intKey] ?? record[name]
+}
+
+function parseBmffMerkleMap(payload: Uint8Array): BmffMerkleMapSegment | null {
+	let decoded: unknown
+	try {
+		decoded = decode(payload)
```
littlespex: §A.5.4.1.4 allows trailing padding bytes after the CBOR (and requires fixed-size padding when multiple merkle boxes for one tree share a file). `cbor-x`'s `decode()` throws `Data read, but end of buffer not reached` on trailing bytes (verified locally), so a spec-conforming padded box lands in this `catch` and gets flagged `ASSERTION_BMFFHASH_MALFORMED`. The same applies to the `decode` in `readSpecAuxPayload`. Suggest decoding tolerantly (read one CBOR item, ignore the tail) and adding a padded-box test.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,297 @@
...
+			codes.add(C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED)
+			continue
+		}
+
+		if (location === null) location = segmentMap.location
+
+		// §15.12.2: each track's location must increment by 1 during sequential playback.
+		const trackKey = `${segmentMap.uniqueId}:${segmentMap.localId}`
+		const lastLocation = previousLocations.get(trackKey)
+		const locationValid = lastLocation === undefined || segmentMap.location === lastLocation + 1
```
littlespex, line 262: Two notes on the continuity check:

- §15.12.2 says locations for a given Merkle tree "start at zero and increments by one". Accepting any first location when there's no prior state is a reasonable trade-off for seek support, but a caller who wants strict start-at-zero can't express it: the `uniqueId:localId` key format of `MerkleSegmentState.lastLocation` isn't documented, so they can't seed the state. Worth documenting the key format or noting the leniency in the TSDoc.
- `LiveVideoStatusCode.ASSERTION_INVALID` is documented as "Live video assertion field invalid: sequenceNumber or streamId mismatch (§19.7.2)", and this is a VOD path. At minimum that enum TSDoc needs updating; a dedicated discontinuity code might fit better.

libs/c2pa/src/merkle/validateMerkleMaps.ts
```diff
@@ -0,0 +1,92 @@
...
+
+	const merkleMaps = extractMerkleMaps(data as Record<string, unknown>)
+	if (!merkleMaps) {
+		codes.add(C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED)
+		return []
+	}
+
+	const initHashByAlg = new Map<string | null, Uint8Array>()
+	for (const merkleMap of merkleMaps) {
+		if (!merkleMap.initHash) continue
```
littlespex, line 82: Per the `merkle-map` CDDL, `initHash` is *required* for fMP4 assets split across multiple files ("this field is required to be present and is the hash of the entire initialization segment file…"), which is exactly this API's use case. Skipping silently means a merkle-only init segment whose maps omit `initHash` has nothing binding the init bytes (moov, codec config), yet `validateC2paInitSegment` still returns `isValid: true`. Suggest flagging `ASSERTION_BMFFHASH_MALFORMED` when `initHash` is absent, or documenting why it's tolerated.

libs/c2pa/src/merkle/validateMerkleMaps.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (rawInitHash != null && !initHash) return null
+
+		const rawAlg = record['alg'] ?? assertionAlg
+		maps.push({
+			uniqueId,
+			localId,
+			count,
+			hashes,
+			initHash,
+			alg: typeof rawAlg === 'string' ? normalizeAlgorithmName(rawAlg) : null,
```
littlespex: Spec nit: per the CDDL, if `alg` is absent from the merkle-map, the enclosing assertion, and the claim, "this structure is invalid; there is no default". Here a missing alg flows through as `null` and `verifyMerkleProof` defaults to SHA-256 via `normalizeAlgorithmName`. That matches the lib-wide default so it may be intentional, but it deserves a comment.

libs/c2pa/src/merkle/MerkleSegmentValidation.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @public
+ */
+export type MerkleMap = {
+	readonly uniqueId: number
+	readonly localId: number
+	readonly count: number
+	readonly hashes: readonly Uint8Array[]
+	readonly initHash: Uint8Array | null
+	readonly alg: string | null
+	readonly exclusions: readonly {
```
littlespex: This inlines the shape of `BmffHashExclusion`/`BmffHashConstraint` (which already carry `@public` TSDoc but aren't exported from the barrel). Exporting those types and referencing them here would keep the two shapes from drifting.

libs/c2pa/src/merkle/MerkleSegmentValidation.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Continuity state carried between calls to `validateC2paMerkleSegment`.
+ * Reset (pass `undefined`) after a seek.
+ *
+ * @public
+ */
+export type MerkleSegmentState = {
+	readonly lastLocation: ReadonlyMap<string, number>
```
littlespex: Naming nit: it's a map of locations, so `lastLocations`. Also, the `${uniqueId}:${localId}` key format is now public API surface (callers hold this state and could seed or inspect it) but isn't documented anywhere.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,297 @@
...
+ * Validates a fragmented MP4 VOD media segment against the init manifest's
+ * merkle maps (§15.12.2.2). Pure: the caller persists `nextState` and resets
+ * `state` after a seek. Returns `null` when `merkleMaps` is empty.
+ *
+ * @example
+ * {@includeCode ../../test/merkle/validateC2paMerkleSegment.test.ts#example}
+ *
+ * @public
+ */
+export async function validateC2paMerkleSegment(
```
littlespex, line 224: Ergonomics nit, not blocking: this is the only `validateC2pa*` function returning `| null`. The caller already holds `merkleMaps` and can gate on `length > 0`, so a non-null return matching the other validators would remove a narrowing step from every call site.

littlespex (commented):
Reviewed against the C2PA 2.3 and 2.4 spec text (cross-checked section by section) and c2pa-rs as the interop reference. Also ran the suite locally: 146/146 pass, build and API Extractor clean.

The core is in good shape: the proof walk matches the §18.6.6.1 worked example, the offset-prefix heuristic matches §18.6.2 exactly, and the state semantics (no baseline advance on failure, reset-on-seek) are well designed and well tested. The proof-path and discontinuity test suites are thorough.

Main threads in the inline comments:

1. **Aux-box format provenance.** The "spec §A.5.4 format" branch (`MERKLE_AUX_UUID` + CBOR `{box_purpose, data}` + integer keys) doesn't appear in C2PA 2.3, 2.4, or any implementation I could find. The "JUMBF fallback" is actually the spec's format. Needs a citation or removal — this is the one thing I'd like resolved before merge.
2. **Robustness against conforming content:** trailing padding after the aux-box CBOR makes `cbor-x` throw (flagged malformed), and a missing `initHash` (required by the CDDL for multi-file fMP4) is accepted silently.
3. **Docs** (files not in this diff, so noting here): README still says "both segment validation methods" and doesn't mention the new API; there's no `docs/merkle-validation.md` alongside the VSI and manifest-box pages; `docs/results-and-error-codes.md` lacks the two new codes and the `SESSIONKEY_INVALID` behavior change.

libs/c2pa/src/C2paStatusCode.ts
```diff
@@ -19,6 +19,10 @@ export const C2paStatusCode = {
 	ASSERTION_ACTION_INGREDIENT_MISMATCH: 'assertion.action.ingredientMismatch',
 	/** Claim signature verification failed (§15.7) */
 	CLAIM_SIGNATURE_MISMATCH: 'claim.signature.mismatch',
+	/** BMFF hash assertion or Merkle structure is malformed (§D.3) */
```
littlespex: I couldn't locate a §D.3 in the 2.4 spec — these codes are defined in the §15.12.2 validation text. Can you fix the citation (here and on line 24)? Related: `docs/results-and-error-codes.md` should gain these two codes and note the `SESSIONKEY_INVALID` behavior change.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: This assumed a second box format existed; the spec only defines one ContentProvenanceBox, reused for every box_purpose. Fixed in 5046af2, kept the version/flags + purpose-string format.

libs/c2pa/src/utils.ts
valentinamgiusti: Same mistake as the sibling comment: this UUID was created under that assumption. Fixed in 5046af2.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: The CDDL was backwards here: string keys are the only ones defined, not a fallback. Fixed in 5046af2.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: cbor-x's decode() throws on trailing bytes, so a spec-valid padded box was rejected as malformed. Fixed in 9ecd69b via decodeMultiple, plus a padded-box test.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: Both missing: the key format was public API with no docs, and the ASSERTION_INVALID reuse wasn't documented. Fixed in daefac3.

libs/c2pa/src/merkle/validateMerkleMaps.ts
valentinamgiusti: initHash was treated as optional, but the CDDL requires it for fragmented assets; missing it left the init segment unbound. Fixed in 9f267a4.

libs/c2pa/src/merkle/validateMerkleMaps.ts
valentinamgiusti: The CDDL disallows a default alg here; this was falling back to SHA-256 silently. Fixed in 9f267a4 (same commit).

libs/c2pa/src/merkle/MerkleSegmentValidation.ts
valentinamgiusti: No reason to keep an inline duplicate once the real type was public. Fixed in ee5dd54.

libs/c2pa/src/merkle/MerkleSegmentValidation.ts
valentinamgiusti: It's a map so the plural fits, and the key format needed documenting. Fixed in daefac3.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: The null case was never reachable given how the one caller uses this. Fixed in 5320ebd.

libs/c2pa/src/C2paStatusCode.ts
valentinamgiusti: Wrong annex: this package targets 2.3, which has no Annex D. Fixed in 7e70d8d, citing §15.2.2.3. Might need a separate PR later to move citations to 2.4.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,291 @@
...
+import type { UuidParsedBox } from '../utils.ts'
+import type { MerkleMap, MerkleSegmentState, MerkleSegmentValidation } from './MerkleSegmentValidation.ts'
+
+type MerkleValidationCode = LiveVideoStatusCode | C2paStatusCode
+
+const MERKLE_BOX_PURPOSE = 'merkle'
+
+// --- Auxiliary uuid box extraction ---
+
+type AuxPayload = Uint8Array | 'other' | 'malformed'
```
littlespex: Non-blocking cleanup: after 5046af2, `readAuxPayload` can only return a payload or `'other'`, so the `'malformed'` variant here and the `malformed` flag plumbing in `extractMerkleAuxBoxes` (lines 25-42) are dead code — the flag is always false. `readAuxPayload` could return `Uint8Array | null` and the extractor could return just the payloads array.

libs/c2pa/src/merkle/validateMerkleMaps.ts
```diff
@@ -0,0 +1,97 @@
...
+
+		// §A.5.4.2/merkle-map CDDL: initHash is required for fragmented assets, which
+		// is the only case this function handles (it's only called for init segments).
+		const initHash = toUint8Array(record['initHash'])
+		if (!initHash) return null
+
+		// §A.5.4.2/merkle-map CDDL: no default when alg is absent from both this
+		// entry and the enclosing assertion; the structure is invalid.
+		const rawAlg = record['alg'] ?? assertionAlg
+		if (typeof rawAlg !== 'string') return null
```
littlespex, line 49: Non-blocking spec pedantry: the CDDL's fallback chain for a missing merkle-map `alg` goes entry → enclosing bmff-hash-map → claim-level `alg` (§15.4 in 2.4 numbering), and only then is the structure invalid. This check stops at the assertion level, so an asset that only sets the claim-level alg would be rejected as malformed. c2pa-rs always writes assertion-level alg, so this is theoretical — worth at most a comment noting the claim-level fallback isn't plumbed through here.

littlespex (commented):
Verified all the fix commits against the spec and cbor-x behavior — every thread from the first round is properly addressed, and the continuity resync in d60bbf6 is a genuine improvement (only proof-verified content can advance the window; tampered, replayed, and reordered segments are still flagged).

Two tiny non-blocking nits inline, plus one docs item that can't be an inline comment because the files aren't in the diff: the README still describes "both segment validation methods" and doesn't mention `validateC2paMerkleSegment`, and there's no `docs/merkle-validation.md` alongside `vsi-validation.md` and `manifest-box-validation.md`. Fine as a follow-up PR if you'd rather land this one first.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
valentinamgiusti: Dead code since 5046af2 removed the only path that returned 'malformed'. Fixed in 1350757: readAuxPayload now returns Uint8Array | 'other', extractMerkleAuxBoxes just returns the payloads array.

libs/c2pa/src/merkle/validateMerkleMaps.ts
valentinamgiusti: Confirmed c2pa-rs always writes assertion-level alg, so no real content hits the claim-level fallback gap. Added a comment noting it's not plumbed through here in 1350757, instead of threading claim data through this function.

libs/c2pa/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@svta/cml-c2pa",
-	"version": "1.0.1",
+	"version": "1.1.0",
```
littlespex: @valentinamgiusti Revert the version number changes. We have a separate build script that handles this.

package-lock.json
```diff
@@ -53,7 +53,7 @@
 		},
 		"libs/c2pa": {
 			"name": "@svta/cml-c2pa",
-			"version": "1.0.1",
+			"version": "1.1.0",
```
littlespex: @valentinamgiusti Revert the version number changes. We have a separate build script that handles this.

libs/c2pa/package.json
valentinamgiusti: Reverted in c8562fc.

package-lock.json
valentinamgiusti: Reverted in c8562fc (same commit).

littlespex (approved):

littlespex (changes requested):
@valentinamgiusti The `README` still describes "both segment validation methods" and doesn't mention `validateC2paMerkleSegment`, and there's no `docs/merkle-validation.md` alongside `vsi-validation.md` and `manifest-box-validation.md`.

valentinamgiusti:
Fixed in 3a10e3d: reworded the intro to list all three methods instead of "both", added `validateC2paMerkleSegment` to the README's Usage section, and added `docs/merkle-validation.md` following the same structure as `vsi-validation.md` and `manifest-box-validation.md`.

littlespex (approved):

PR #379, opened 2026-07-17
feat(c2pa): support custom continuity methods in manifest-box validation

valentinamgiusti:
## Summary

Adds support for implementer-defined continuity methods (C2PA §19.3.2) in `validateC2paManifestBoxSegment`:

- New `livevideo.continuityMethod.unsupported` status code, emitted alongside the spec-mandated `continuityMethod.invalid` (§19.7.2) so consumers can distinguish an unverifiable custom method from a broken chain.
- Optional `options.continuityValidator: { method, validate }` registers a validator for a single custom method. The callback receives the decoded `c2pa.livevideo.segment` assertion (including its method-specific fields) and the parsed manifest; the built-in `c2pa.manifestId` method cannot be overridden.

Motivated by third-party PSM streams using custom anchor-chain continuity methods, which previously validated as a broken chain with no way to tell why.

## Test plan

- [x] `npm run build -w=libs/c2pa` compiles and API Extractor passes
- [x] `npm test -w=libs/c2pa`: 154/154 tests pass

> **Note:** Stacked on #378 (its commits are included in this diff); only the last 5 commits are new here.


libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,293 @@
...
+
+	const layers = treeLayout(count)
+	if (!layers.includes(manifestRow.length)) {
+		return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
+	}
+
+	const treeDepth = layers.length - 1
+	const path = (proofPath ?? []).filter((node): node is Uint8Array => node !== null)
+	if (path.length > treeDepth) return C2paStatusCode.ASSERTION_BMFFHASH_MALFORMED
+
```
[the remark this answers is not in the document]
valentinamgiusti: Fixed in 742dc93 (on the base branch, #378).

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
```diff
@@ -0,0 +1,293 @@
...
+		}
+
+		const leafHash = await computeLeafHash(segmentBytes, merkleMap, leafHashCache)
+		if (bmffHashHex === null) bmffHashHex = bytesToHex(leafHash)
+		const proofResult = await verifyMerkleProof(leafHash, segmentMap.location, segmentMap.hashes, merkleMap)
+		if (proofResult === 'ok') {
+			nextLocations.set(trackKey, segmentMap.location)
+		} else {
+			codes.add(proofResult)
+		}
```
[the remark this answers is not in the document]
valentinamgiusti: Fixed in 838effc (on the base branch, #378).

valentinamgiusti:
This branch is stacked on #378. Any new commit there requires rebasing this branch on top, which is why it gets force-pushed too. This PR's own commits stay unchanged, only the base moves.

libs/c2pa/src/merkle/validateC2paMerkleSegment.ts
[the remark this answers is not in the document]
valentinamgiusti: Follow-up: freezing the baseline on discontinuity (838effc) caused a worse regression, one dropped segment left everything after it flagged forever. d60bbf6 (also on #378) advances the baseline on any proof-verified segment instead. A replayed segment is still flagged invalid when it happens, it just doesn't stay frozen after.

libs/c2pa/docs/results-and-error-codes.md
```diff
@@ -36,6 +36,7 @@ import { LiveVideoStatusCode } from '@svta/cml-c2pa'
 | `SEGMENT_INVALID` | `livevideo.segment.invalid` | Crypto failure: signature, hash, or key mismatch |
 | `ASSERTION_INVALID` | `livevideo.assertion.invalid` | sequenceNumber or streamId mismatch |
 | `CONTINUITY_METHOD_INVALID` | `livevideo.continuityMethod.invalid` | Continuity chain broken or method unsupported |
```
littlespex: "Continuity chain broken" is stale here: a broken chain emits `SEGMENT_INVALID`, and this PR's update to manifest-box-validation.md now says exactly that, so the two docs contradict each other. `CONTINUITY_METHOD_INVALID` is actually raised when the method is absent or unsupported, or when `previousManifestId` is missing.

```suggestion
| `CONTINUITY_METHOD_INVALID` | `livevideo.continuityMethod.invalid` | `continuityMethod` absent, unsupported, or required companion fields missing |
```

The `SEGMENT_INVALID` row above could also mention continuity failures (broken `c2pa.manifestId` chain, failed custom validator), since this PR adds a second non-crypto source of that code.

libs/c2pa/CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ and this project adheres to
 
 ## [Unreleased]
 
+## [1.1.0] - 2026-07-13
+
+### Added
+
+- VOD Merkle validation (C2PA §15.12.2 / §18.6): `validateC2paMerkleSegment` verifies fragmented MP4 media segments against the merkle maps from the init manifest — per-track Merkle proof verification with `location` continuity enforced via caller-held state
+- Custom continuity method support in manifest-box validation (C2PA §19.3.2 / §19.7.2): `validateC2paManifestBoxSegment` accepts an optional `continuityValidator` to verify implementer-defined continuity methods; segments declaring an unregistered method keep failing with `livevideo.continuityMethod.invalid` per spec
```
littlespex, line 13: Release-coordination note: these entries land under `[1.1.0] - 2026-07-13`, whose release-prep commit lives in #378. If 1.1.0 is published from #378 before this PR merges, the published version won't contain these features (new API surface: the `continuityValidator` option, two exported types, one status code), and this PR would then need its own minor bump. Fine if both PRs merge before publishing; otherwise consider keeping these bullets under `[Unreleased]`. The section date will also need a refresh at publish time.

libs/c2pa/test/manifestbox/validateC2paManifestBoxSegment.test.ts
```diff
@@ -5,6 +5,23 @@ import { describe, it } from 'node:test'
 import { encode } from 'cbor-x/encode'
 import { computeBmffHash } from '../../src/bmff/computeBmffHash.ts'
 
+const CUSTOM_METHOD = 'com.test.custom'
+
+function loadFixture(name: string): Uint8Array {
+	return new Uint8Array(readFileSync(new URL(`../fixtures/${name}`, import.meta.url)))
+}
+
+function withCustomContinuityMethod(bytes: Uint8Array): Uint8Array {
```
littlespex, line 14: Non-blocking coverage note: because this helper rewrites assertion bytes inside the signed fixture, the claim's hashed-URI binding breaks, so no test can reach `isValid: true` for a custom method. Verified empirically: the patched fixture with a passing validator yields exactly `['assertion.hashedURI.mismatch']`. The continuity codes are still meaningfully asserted, but nothing proves a custom-method segment can validate end to end. The unsigned segment builder in the §18.6.2 describe block below could be extended with a `c2pa.livevideo.segment` assertion to build a clean happy-path fixture in a follow-up.

libs/c2pa/src/manifestbox/validateC2paManifestBoxSegment.ts
```diff
@@ -164,19 +132,35 @@ function collectErrorCodes(
...
 
-	return [...codes]
+	const custom = options?.continuityValidator
+	if (custom?.method !== method || !manifest || !liveVideo) {
+		return [LiveVideoStatusCode.CONTINUITY_METHOD_INVALID, LiveVideoStatusCode.CONTINUITY_METHOD_UNSUPPORTED]
+	}
+
+	const ok = await Promise.resolve()
+		.then(() => custom.validate(liveVideo.data, manifest))
+		.catch(() => false)
```
littlespex, line 162: Two minor notes, neither blocking:

- `!manifest || !liveVideo` is runtime-unreachable (a non-null `method` implies `liveVideo` is non-null, which implies `manifest` is non-null), so it's effectively TS narrowing. Fine as is; just noting it's dead as a behavior branch.
- `.catch(() => false)` silently maps validator crashes to `SEGMENT_INVALID`. Documented and fail-closed, which is right for a validation library, but consumers can't distinguish "my validator threw" from "continuity failed". If that ever matters, a debug hook could be added later.

libs/c2pa/docs/manifest-box-validation.md
```diff
@@ -131,9 +131,29 @@ Segment 3: manifestId = "ghi-789", previousManifestId = "def-456"  // must match
 > [!NOTE]
 > Every segment must declare a `previousManifestId`, including the first one. The first segment typically references the initial static manifest. When `lastManifestId` is `null` (first call), the chain comparison is skipped — the validator only checks that `previousManifestId` is present.
 
-The `continuityMethod` field indicates how continuity is verified. The currently supported method is `c2pa.manifestId`.
+The `continuityMethod` field indicates how continuity is verified. The only method validated built-in is the spec-defined `c2pa.manifestId` (C2PA §19.3.2).
```
littlespex: Wording: "The only method validated built-in" reads awkwardly.

```suggestion
The `continuityMethod` field indicates how continuity is verified. The only built-in method is the spec-defined `c2pa.manifestId` (C2PA §19.3.2).
```

libs/c2pa/docs/manifest-box-validation.md
```diff
@@ -131,9 +131,29 @@ Segment 3: manifestId = "ghi-789", previousManifestId = "def-456"  // must match
...
 
-When the chain is broken (the `previousManifestId` does not match the previous segment's manifest ID), the result includes `LiveVideoStatusCode.CONTINUITY_METHOD_INVALID`.
+When the chain is broken (the `previousManifestId` does not match the previous segment's manifest ID), the result includes `LiveVideoStatusCode.SEGMENT_INVALID`.
+
+### Custom continuity methods
+
+The spec allows implementers to define their own continuity methods (§19.3.2), identified by a custom label (e.g. `com.example.anchor-chain`). Since their semantics are implementation-specific, this library cannot validate them by default: segments declaring an unrecognized method fail with `LiveVideoStatusCode.CONTINUITY_METHOD_INVALID`, as required by §19.7.2, plus `LiveVideoStatusCode.CONTINUITY_METHOD_UNSUPPORTED` - lets consumers distinguish an unverifiable method from a broken chain.
```
littlespex: The clause after the hyphen is a sentence fragment.

```suggestion
The spec allows implementers to define their own continuity methods (§19.3.2), identified by a custom label (e.g. `com.example.anchor-chain`). Since their semantics are implementation-specific, this library cannot validate them by default: segments declaring an unrecognized method fail with `LiveVideoStatusCode.CONTINUITY_METHOD_INVALID`, as required by §19.7.2, plus `LiveVideoStatusCode.CONTINUITY_METHOD_UNSUPPORTED`, which lets consumers distinguish an unverifiable method from a broken chain.
```

littlespex (commented):
Reviewed the #379-scope changes (custom continuity methods). The Merkle work belongs to #378, so I kept it out of scope by diffing `pull/378/head..pull/379/head`.

**Verdict: LGTM once the two doc/process comments are addressed.** The feature logic is correct and spec-aligned: fail-closed per §19.7.2 for unregistered methods, the built-in `c2pa.manifestId` method can't be overridden, and the validator verdict can't mask other error codes. The API addition is backwards compatible and the continuity logic is well covered by tests.

Verified locally at 499d20a: built all workspaces, `npm test -w=libs/c2pa` passes 162/162; CI and DCO are green.

Inline: two items to fix (stale error-code table row, changelog/release coordination) and four non-blocking notes (happy-path coverage gap, two code nits, two doc wording suggestions).

libs/c2pa/docs/results-and-error-codes.md
valentinamgiusti: Fixed in 3440a58: corrected the CONTINUITY_METHOD_INVALID row and added continuity as a SEGMENT_INVALID source too.

libs/c2pa/CHANGELOG.md
valentinamgiusti: Moved to Unreleased in 3440a58, agreed it shouldn't assume merge order relative to #378's release.

libs/c2pa/test/manifestbox/validateC2paManifestBoxSegment.test.ts
valentinamgiusti: Added in f090e4d: an unsigned segment with a matching live-video assertion and flat hash, so a custom method now validates isValid=true end to end instead of only asserting individual codes.

libs/c2pa/src/manifestbox/validateC2paManifestBoxSegment.ts
valentinamgiusti: Agreed on both, leaving as is.

libs/c2pa/docs/manifest-box-validation.md
valentinamgiusti: Fixed in 35b0b80 (folded into the original doc commit).

libs/c2pa/docs/manifest-box-validation.md
valentinamgiusti: Fixed in 35b0b80 (folded into the original doc commit).

libs/c2pa/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@svta/cml-c2pa",
-	"version": "1.0.1",
+	"version": "1.1.0",
```
littlespex: @valentinamgiusti Revert the version number changes. We have a separate build script that handles this.

package-lock.json
```diff
@@ -53,7 +53,7 @@
 		},
 		"libs/c2pa": {
 			"name": "@svta/cml-c2pa",
-			"version": "1.0.1",
+			"version": "1.1.0",
```
littlespex: @valentinamgiusti Revert the version number changes. We have a separate build script that handles this.

libs/c2pa/package.json
valentinamgiusti: Reverted in c8562fc (on the base branch, #378).

package-lock.json
valentinamgiusti: Reverted in c8562fc (on the base branch, #378).

PR #380, opened 2026-07-20
fix(c2pa): enforce 8-byte offset prefix for the flat v3 init hash

littlespex:
## Summary

`validateC2paInitSegment` verified the flat `c2pa.hash.bmff.v3` assertion hash through the lenient dual-mode `validateBmffHash` helper, which tries the 8-byte box-offset prefix first and then falls back to no prefix. C2PA 2.4 §18.6.2 is unambiguous for v2/v3 assertions: the hash input for every non-excluded root box is `offset || data` (8-byte big-endian absolute file offset), and the offset is only omitted for Merkle tree hashes when the bmff-hash-map carries both `hash` and `merkle`. The flat hash field always uses offsets, precisely so "a root box included in the hash cannot change positions in the file."

`validateBmffHashAssertion` now computes the hash strictly with `offsetPrefixSize: 8` and compares, instead of accepting either mode. This also aligns the flat-hash path with the Merkle path introduced in #378, which already derives the prefix mode per the spec rule.

## Evidence

- **Spec**: §18.6.2 (quoted above) mandates the offset prefix for the flat v2/v3 hash with no exception.
- **c2pa-rs interop reference**: verification injects an 8-byte big-endian offset entry per non-excluded top-level box when `bmff_version > 1` (`HashRange::set_bmff_offset` in `bmff_io.rs`, hashed via `start.to_be_bytes()` in `hash_utils.rs`). Strict, no fallback mode.
- **Fixtures/encoders**: both real signed fixtures in the repo (`init_signed_with_session_keys.m4s` and `test-segment.m4s`) match only the prefixed mode, so nothing relies on the unprefixed fallback. The fallback shipped with the original helper in #337 before any caller existed and no issue/PR documents a real unprefixed signer.

The dual-mode `validateBmffHash` helper is intentionally unchanged for the VSI (§19.7.3) and manifest-box (§19.7.2) live-segment paths, where its doc comment documents the inter-signer leniency for the draft live workflow.

## Impact

Before, an init segment whose flat assertion hash was computed without offset prefixes (spec-non-compliant signer) validated cleanly; such hashes do not bind box positions, so excluded-box insertion/resizing went undetected for those streams. Now they fail with `INIT_INVALID`, matching what c2pa-rs would report. Spec-compliant streams are unaffected (the prefixed mode was already tried first).

## Test plan

- New regression tests in `validateC2paInitSegment.test.ts`: prefixed flat hash accepted, unprefixed flat hash rejected (fails on the previous code), and the real signed fixture end-to-end.
- `npm test -w=libs/c2pa`: 124/124 pass. Lint, typecheck, and build clean.

## Requirements Checklist

- [x] Docs/guides updated (no public API change; internal comment cites §18.6.2)
- [x] Changelog updated

### DIFF
--- a/libs/c2pa/CHANGELOG.md
+++ b/libs/c2pa/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- `validateC2paInitSegment` now enforces the 8-byte box-offset prefix (C2PA §18.6.2) when verifying the flat `c2pa.hash.bmff.v3` assertion hash, matching c2pa-rs; unprefixed flat hashes are no longer accepted
+
 ## [1.0.1] - 2026-05-13
 
 ### Changed
--- a/libs/c2pa/src/init/validateC2paInitSegment.ts
+++ b/libs/c2pa/src/init/validateC2paInitSegment.ts
@@ -6,13 +6,13 @@ import type { C2paStatusCode } from '../C2paStatusCode.ts'
 import { LiveVideoStatusCode } from '../LiveVideoStatusCode.ts'
 import { readC2paManifest } from '../readC2paManifest.ts'
 import { extractCertificateFromSignatureBytes } from '../extractManifestCertificate.ts'
-import { validateBmffHash } from '../bmff/validateBmffHash.ts'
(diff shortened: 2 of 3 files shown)


littlespex:
@valentinamgiusti

littlespex:
@N1Knight

PR #381, opened 2026-07-20
fix(c2pa): enforce 8-byte offset prefix for the manifest-box v3 hash

littlespex:
## Summary

- `validateC2paManifestBoxSegment` verified the flat `c2pa.hash.bmff.v3` assertion hash of a §19.7.2 live segment via the lenient dual-mode `validateBmffHash` helper, so unprefixed (spec-noncompliant) flat hashes validated cleanly even though they do not bind box positions. C2PA 2.4 §18.6.2 mandates the 8-byte big-endian box-offset prefix for flat v2/v3 assertion hashes; only Merkle tree hashes may omit it.
- The hash is now computed strictly with `offsetPrefixSize: 8` and compared, matching c2pa-rs and the init-segment change from #380. Unprefixed flat hashes now fail with `SEGMENT_INVALID`.
- The VSI path (§19.7.3) intentionally keeps dual-mode validation: its hash comes from the VSI CBOR map, where the spec does not signal a prefix mode. The `validateBmffHash` doc comment now scopes the inter-signer leniency rationale to VSI-sourced hashes.
- The real signed fixture `test-segment.m4s` matches only the prefixed mode (now covered by an end-to-end regression test), so spec-compliant streams are unaffected.

## Packages Changed

| @svta/cml-c2pa | 1.0.1 | fix |

## Test Plan

- [x] New regression tests in `validateC2paManifestBoxSegment.test.ts`: prefixed flat hash accepted, unprefixed flat hash rejected (fails on the previous code), and the real signed `test-segment.m4s` fixture end-to-end
- [x] `npm test -w libs/c2pa`: 124/124 pass
- [x] Lint, typecheck, and docs build clean (0 errors)

## Requirements Checklist

- [x] Docs/guides updated (no public API change; internal comment cites §18.6.2, helper TSDoc scoped to VSI)
- [x] Changelog updated

### DIFF
--- a/libs/c2pa/CHANGELOG.md
+++ b/libs/c2pa/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- `validateC2paManifestBoxSegment` now enforces the 8-byte box-offset prefix (C2PA §18.6.2) when verifying the flat `c2pa.hash.bmff.v3` assertion hash, matching c2pa-rs; unprefixed flat hashes are no longer accepted. The VSI path (§19.7.3) keeps dual-mode validation since its hash comes from the VSI map, not a §18.6.2 assertion.
+
 ## [1.0.1] - 2026-05-13
 
 ### Changed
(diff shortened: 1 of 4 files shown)


littlespex:
@valentinamgiusti

littlespex:
@N1Knight

PR #385, opened 2026-07-21
Potential fix for code scanning alert no. 3: Insecure randomness

littlespex:
Potential fix for [https://github.com/streaming-video-technology-alliance/common-media-library/security/code-scanning/3](https://github.com/streaming-video-technology-alliance/common-media-library/security/code-scanning/3)

Use a cryptographically secure RNG in the final fallback UUID construction instead of `Math.random()`.  
Best fix without changing functionality: keep the same UUID v4 format logic, but replace random nibble generation with secure random bytes from `crypto.getRandomValues` (browser/Web Crypto API). This preserves output format and removes predictable randomness.

In `libs/utils/src/uuid.ts`, update only the innermost `catch` block (currently lines 20–26). Remove the timestamp/`Math.random()` approach and generate a 16-byte array with `crypto.getRandomValues(new Uint8Array(16))`, then set RFC4122 v4/version and variant bits, and format bytes into canonical UUID string.

No new imports are required if `crypto` is already available globally in this code context (as implied by existing `crypto.randomUUID()` use).

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/libs/utils/CHANGELOG.md
+++ b/libs/utils/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Security
+
+- Replace `Math.random()`-based UUID fallback with `crypto.getRandomValues` to use a cryptographically secure RNG ([#3](https://github.com/streaming-video-technology-alliance/common-media-library/security/code-scanning/3))
+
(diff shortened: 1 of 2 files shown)


littlespex:
@copilot This change needs a CHANGELOG entry

PR #390, opened 2026-07-21
[RFC] Report Transforms for CmcdReporter

littlespex:
## Summary

Proposes an optional `transform` function on the request-report config (top level) and on each event target config: synchronous functions that can customize CMCD report data or cancel reports entirely. Placement determines scope, mirroring `enabledKeys`: the top-level transform governs request-mode reports (`createRequestReport()`), each target's transform governs that target's event reports.

Full proposal: [rfc/cmcd-reporter-middleware.md](https://github.com/streaming-video-technology-alliance/common-media-library/blob/rfc/cmcd-reporter-middleware/rfc/cmcd-reporter-middleware.md)

```ts
	enabledKeys: ['br', 'bl', 'sid', 'cid', 'v', 'sn'],
	// Request mode: never decorate license requests with CMCD
	transform: (data, request) => request.url.includes('/license') ? null : data,
	eventTargets: [{
		url: 'https://collector.example.com/cmcd',
		events: [CmcdEventType.PLAY_STATE, CmcdEventType.RESPONSE_RECEIVED],
		// This target only: drop rr reports for non-segment responses
		transform: (data, request) => {
			if (data.e === CmcdEventType.RESPONSE_RECEIVED && request?.customData?.requestType !== 'segment') {
				return null
			}
			return data
		},
	}],
})
```

> **Revision (2026-07-22)**: the first draft proposed a reporter-level `middleware` array spanning both reporting paths. Review feedback on per-target `rr` filtering (thanks @dsilhavy), plus the observation that `enabledKeys` already establishes a per-placement convention with no global key, reshaped the proposal into per-placement single transform functions. The RFC's Rationale and Revision history sections document the full reasoning.

## Motivation (short version)

`CmcdReporter` has no interception point between data assembly and wire encoding. The `requester` argument runs after batching and encoding (so customizing data means re-parsing, and cancelling a single event inside a batch is impossible), and it never sees request-mode reports at all. Integrators need per-report, per-destination control: redacting fields for third-party collectors, skipping CMCD on non-media requests, sampling periodic reports, and per-target response filtering by request type (the dash.js case, which today forces one reporter instance per target).

## Feedback wanted

The RFC's **Unresolved questions** section lists where community input would help most:

1. Error stance: propagate transform exceptions (proposed) vs fail-closed on throw
2. Should public `recordEvent()` accept an associated `HttpRequest` so custom events can carry request context into transforms?

## Also in this PR

This is the first RFC under a new `rfc/` directory convention (see `rfc/README.md`, adapted from [videojs/v10](https://github.com/videojs/v10/tree/main/rfc)): YAML frontmatter status tracking, PR-based review, and a Final Decision section recorded on acceptance. `AGENTS.md` is updated to split community-facing RFCs (`rfc/`) from implementation plans (`plans/`).


dsilhavy:
Thanks for the quick suggestion @littlespex . One discussion point: 

dash.js has a player-specific filter for `rr` events: a reporting target can opt in only to responses for selected request types, such as `segment` or `mpd`. We currently need separate CmcdReporter instances per target so we can decide whether to call `recordResponseReceived()` for that target.

A global middleware can make this work when target URLs are unique. However, two configured targets may intentionally use the same collector URL but have different event sets, keys, batch sizes, or response filters:

````js
    eventTargets: [
        {
            url: 'https://collector.example/cmcd',
            events: ['rr'],
            enabledKeys: ['e', 'ts', 'url', 'rc', 'sn'],
            batchSize: 1,
            // dash.js extension: only segment responses
            sendResponseReceivedForRequestTypes: ['segment'],
        },
        {
            url: 'https://collector.example/cmcd',
            events: ['ps', 'rr'],
            enabledKeys: ['e', 'ts', 'sta', 'url', 'rc', 'sn'],
            batchSize: 10,
            // dash.js extension: only manifest responses
            sendResponseReceivedForRequestTypes: ['mpd'],
        },
    ],
})
````

Would it make sense to support optional middleware directly on `CmcdEventReportConfig`?
````js
    eventTargets: [{
        url: 'https://collector.example/cmcd',
        events: ['ps', 'rr'],
        middleware: [
            (data, ctx) => {
                if (data.e !== CmcdEventType.RESPONSE_RECEIVED) {
                    return data
                }

                const type = ctx.request?.customData?.request?.type
                return type === 'segment' ? data : null
            }
        ]
    }]
})
````

For dash.js, this would allow one CmcdReporter with all configured event targets. We could keep a global middleware that runs first and the target middleware runs second.

**Comment by littlespex** [comment]
@dsilhavy. This simplifies things. We can drop the global middleware array entirely and just use a single optional transform function per target, scoped the same way `enabledKeys` is: top-level for request-mode reports or per event target.

Your example becomes:

```js
	eventTargets: [{
		url: 'https://collector.example/cmcd',
		events: ['ps', 'rr'],
		transform: (data, request) => {
			if (data.e !== 'rr') return data
			return request?.customData?.request?.type === 'segment' ? data : null
		},
	}],
})
```

The transform lives on the target it governs, so same-URL targets are a non-issue, and request-type taxonomy stays in dash.js.

Updated RFC: https://github.com/streaming-video-technology-alliance/common-media-library/blob/rfc/cmcd-reporter-middleware/rfc/cmcd-reporter-middleware.md

One question: does a single function per target (instead of an array) still work for dash.js? Composing multiple concerns in one target would be plain function composition.

dsilhavy:
> @dsilhavy. This simplifies things. We can drop the global middleware array entirely and just use a single optional transform function per target, scoped the same way `enabledKeys` is: top-level for request-mode reports or per event target.
> 
> Your example becomes:
> 
> ```js
> const reporter = new CmcdReporter({
> 	eventTargets: [{
> 		url: 'https://collector.example/cmcd',
> 		events: ['ps', 'rr'],
> 		transform: (data, request) => {
> 			if (data.e !== 'rr') return data
> 			return request?.customData?.request?.type === 'segment' ? data : null
> 		},
> 	}],
> })
> ```
> 
> The transform lives on the target it governs, so same-URL targets are a non-issue, and request-type taxonomy stays in dash.js.
> 
> Updated RFC: https://github.com/streaming-video-technology-alliance/common-media-library/blob/rfc/cmcd-reporter-middleware/rfc/cmcd-reporter-middleware.md
> 
> One question: does a single function per target (instead of an array) still work for dash.js? Composing multiple concerns in one target would be plain function composition.

Thanks @littlespex looks good to me. A single function per target is sufficient to handle the `sendResponseReceivedForRequestTypes` logic in dash.js. If additional target-specific rules arise, we can compose them in one function or use a small local transform-composition helper

rfc/cmcd-reporter-middleware.md
```diff
@@ -0,0 +1,323 @@
...
+- Transform-added keys remain subject to `enabledKeys`.
+- Error propagation from both paths.
+- A `#region` example block wired into the TSDoc via `{@includeCode}`.
+
+## Drawbacks
+
+- **Cross-cutting policy has no single registration point.** A concern that applies everywhere must be referenced from the top level and from each target. Mitigations: shared function references (see Guide-level), and the observation that whole-session data minimization is already served by `enabledKeys` placement and by not feeding sensitive values into `update()`.
+- **One function per placement.** Integrators composing several concerns at one placement write the composition themselves. A `composeTransforms()` helper can ship later as a separate tree-shakeable export if demand appears.
+- **Footgun potential**: a transform can degrade payloads (e.g. stripping keys a collector expects). The reporter-owned-field re-stamping and `enabledKeys` filtering bound the damage to data quality; structural validity and sequence integrity are protected.
```
littlespex: **[P1] Preserve required event fields after transforms**

Only `e`, `sn`, and `msd` are protected after transformation, so a transform can remove `sta`, `pr`, `cid`, `bg`, `br`, `cen`, `ec`, or `url` and emit an invalid event. I verified that removing `sta` produces `e=ps,ts=1,v=2`, which `validateCmcdEvents()` rejects. Preserve or re-stamp every event-required field, or cancel/reject invalid transformed output, and cover each required-field family in tests.

rfc/cmcd-reporter-middleware.md
```diff
@@ -0,0 +1,323 @@
...
+
+A transform is a pure synchronous function:
+
+```ts
+(data: Cmcd, request: HttpRequest | undefined) => Cmcd | null
+```
+
+- Return the data object (the same one, mutated in place if you like, or a copy) to send it.
+- Return `null` to cancel the report. Nothing is sent.
+- The `data` argument is a per-report copy, so mutating it never affects the reporter's persistent state or other targets' reports.
```
littlespex: **[P1] Do not promise isolation from a shallow copy**

The proposed `{ ...persistentData, ...eventData }` construction is shallow, while `Cmcd` contains arrays and nested `SfItem` values. An allowed in-place mutation such as `data.ec.push(...)` changes the persistent or caller-owned array, and the next target sees the mutation. Either deep-clone nested values, with the associated performance cost, or make nested mutation unsupported through a deeply readonly contract and require callers to return copied nested values.

rfc/cmcd-reporter-middleware.md
```diff
@@ -0,0 +1,323 @@
...
+Transforms run before encoding, so the `enabledKeys` allowlist (and the encoder's force-include rules for required event fields) still applies to transform output:
+
+- Keys added by a transform must be present in the `enabledKeys` at the same placement to reach the wire.
+- Keys removed by a transform stay removed.
+
+This keeps `enabledKeys` as the single wire allowlist and prevents a transform from accidentally leaking keys a target was never configured to receive. Note that data minimization is already a per-placement concern in this API: a key that must never reach any destination should simply not be enabled anywhere, which is also why a both-paths global hook is unnecessary for that job.
+
+### Error handling
+
+Transform exceptions propagate to the caller: `createRequestReport()`, `recordEvent()`, `update()`, or, for `TIME_INTERVAL` events, the interval timer context. The library does not swallow them because both silent alternatives are worse:
```
littlespex, line 275: **[P1] Make transform failures leave state retryable**

`recordEvent()` commits `lastEmitted` before target fan-out. If a target transform throws, iteration and `processEventTargets()` abort; retrying the same state is then suppressed by dedup, and later targets never receive the transition. Injecting a throw at `recordTargetEvent()` and retrying `update({ sta: 'p' })` produced zero reports. Define transactional staging or rollback/per-target isolation before adopting exception propagation.

rfc/cmcd-reporter-middleware.md
```diff
@@ -0,0 +1,323 @@
...
+> into per-placement `transform` functions. See Rationale and alternatives for
+> why. The file name retains the original working name.
+
+## Summary
+
+Add an optional `transform` function to the request-report configuration and to each event target configuration. A transform is a synchronous function that receives the assembled CMCD data for a single report, plus the associated media request when one exists, and returns either the (possibly modified) data to continue or `null` to cancel the report.
+
+Placement determines scope, mirroring how `enabledKeys` already works: the top-level `transform` applies to request-mode reports produced by `createRequestReport()`, and each event target's `transform` applies to event reports bound for that target.
+
+```ts
```
littlespex, line 27: **[P2] Make RFC examples complete and runnable**

The repository requires every code example to be complete and runnable, but this example omits imports and later snippets reference undeclared symbols such as `Cmcd` and `scrubPii`. Add the necessary imports and definitions so adopters can typecheck and run each example directly.

rfc/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## What Belongs Here
+
+- Changes to the public API surface of any package
+- New packages or significant changes to core architecture
+- Spec-interpretation decisions that affect wire behavior
+- Anything that needs alignment from adopters or maintainers before implementation
+
+**Use a plan instead** (`plans/<feature-name>/`) for implementation details and design-session artifacts you own.
+
```
littlespex, line 13: **[P2] Resolve the small-public-feature ambiguity**

This section says all public API changes belong in an RFC, but immediately says small features should skip RFCs and plans. A small exported option or function satisfies both rules. Qualify the exception as non-public/internal work, or explicitly state that the public-API rule takes precedence.

littlespex (commented):
Five findings from the requested RFC review: three contract-level correctness issues and two documentation/process issues.

littlespex:
## v4: three contract amendments

Pushed `f6dccc1`. All five review findings are addressed. Three of them were contract-level, so the RFC now specifies behavior it previously either got wrong or left open. Each was reproduced against a working implementation before and after the fix, so these are not speculative.

**1. Transforms can no longer strip keys the event requires.** v3 protected only `e`, `sn`, and `msd`. A transform returning `{ ...data, sta: undefined }` on a `ps` event emitted:

```
cid="cid-1",e=ps,sid="sid-1",sn=0,ts=1785170848124,v=2
```

which this library's own `validateCmcdEventReport()` rejects with `State-change event (e="ps") requires the "sta" key to be present`. The reporter now captures the event's required keys before the transform and restores anything the transform removed. Required keys are `e` and `ts` for every event, the signalled field for state-change events (`sta`, `pr`, `cid`, `bg`, `br`), `cen` for `ce`, `ec` for the error event, and `url` for `rr`. Note that `ts` was missing from the original finding's list and is equally strippable.

Restoration deliberately does not fabricate: a required key already absent before the transform ran stays absent, because that is a bug at the call site rather than something the transform did. The v3 Drawbacks section claimed "structural validity and sequence integrity are protected", which was simply false; it is now accurate.

**2. Nested values in the report copy are copied.** v3 promised that mutating the `data` argument never affects the reporter's persistent state or other targets' reports, but the copy was a shallow spread and 16 CMCD keys are array-typed. With target A's transform doing `data.ec.push('LEAKED')`, target B, which had no transform at all, emitted the pushed value, and the reporter's persistent array kept growing on every later report.

Two findings from implementing this that changed the approach. `SfItem` is a class, and `prepareCmcdData`, `CMCD_FORMATTER_MAP`, `validateCmcdValues`, and the structured-field encoder all branch on `instanceof SfItem`, so the copy must preserve prototypes. `structuredClone` does not preserve them, so it would have silently turned every `SfItem` into a plain object and changed what goes on the wire. It is now rejected on correctness first and cost second.

The copy is complete rather than best-effort because the value space is bounded: `CmcdValue` and `CmcdCustomValue` admit only primitives, `SfItem<primitive>`, and arrays of those, with a flat `params` record. Measured cost is roughly 200ns for the copy in isolation and roughly 400ns end to end including the transform call, against a roughly 8.5µs report path, so about 5%, and only where a transform is configured.

**3. Event-mode exception propagation is isolated per target.** This was the most severe of the three. `recordEvent()` commits the state-change dedup baseline before fanning out to targets, and that baseline is never rolled back. In v3 an unqualified throw aborted the fan-out, so targets ordered after the throwing one never received the transition, and the player's retry of the same state was suppressed by dedup. Because target order is fixed by configuration, one consistently throwing transform starved every later target for the rest of the session, not just for one event. Reproduced: first `update({ sta: 'p' })` threw, the retry did not even throw because dedup suppressed it, and a later distinct `update({ sta: 'a' })` threw again with the healthy target still at zero reports.

Each target's transform is now invoked in isolation. A throw drops that target's report, the remaining targets are still processed, queued batches are still sent, and the first error is re-thrown after the fan-out completes. The throwing target consumes no sequence number, exactly as if its transform had returned `null`. Propagation as the error stance is unchanged; only the blast radius is. I also dropped v3's promise to attach additional target errors as `cause`, since chaining integrator error objects is not worth an aggregate error type for a condition the contract already forbids.

The two documentation findings are fixed as well: RFC examples are self-contained with imports and a defined `scrubPii`, and `rfc/README.md` now states that the public-API rule takes precedence over size, with "not public" as the exception rather than "small".

@dsilhavy this is why I have not leaned on your earlier sign-off to merge. The API shape is unchanged and the per-placement single-function design you confirmed is untouched, so nothing here revisits the decision you reviewed. But amendment 3 does change observable error behavior from what you approved, so a fresh look at the Error handling and Report data copying sections would be worth having before this merges. All three amendments are strictly more conservative than v3: they remove ways to emit an invalid report, to corrupt another target's data, and to silently lose reports.

rfc/cmcd-reporter-middleware.md
littlespex: Fixed in f6dccc1. Reproduced first: a transform returning `{ ...data, sta: undefined }` on a `ps` event emitted `cid="cid-1",e=ps,sid="sid-1",sn=0,ts=...,v=2`, which `validateCmcdEventReport()` rejects exactly as you said.

The reporter now captures the event's required keys before the transform and restores whatever the transform removed. Adding to your list: `ts` belongs on it too, since every event requires it and it is equally strippable. Full set is `e` and `ts` for all events, the signalled field for state-change events, `cen` for `ce`, `ec` for the error event, `url` for `rr`.

I went with restore rather than cancel-on-invalid, because `prepareCmcdData` already force-includes the required state field after the `enabledKeys` filter, so the library already treats these keys as non-negotiable. Restoration does not fabricate: a required key absent before the transform stays absent, since that is a call-site bug. Also corrected the Drawbacks claim that structural validity was already protected. Covered by one test per required-key family, each asserting the result against `validateCmcdEventReport()`.

rfc/cmcd-reporter-middleware.md
littlespex: Fixed in f6dccc1, and it was worse than described. With target A doing `data.ec.push(...)`, target B, which had no transform at all, emitted the pushed value, and the persistent array kept growing on every later report, so it corrupted the reporter's own state rather than just leaking across targets.

I took the clone side of your either/or, but not a general deep clone. Two things surfaced while implementing it. `SfItem` is a class and `prepareCmcdData`, `CMCD_FORMATTER_MAP`, `validateCmcdValues`, and the encoder all branch on `instanceof SfItem`, so the copy has to preserve prototypes. `structuredClone` does not, so it would have silently turned every `SfItem` into a plain object and changed wire output; it is now rejected on correctness before cost.

The copy is complete rather than best-effort because `CmcdValue` and `CmcdCustomValue` bound the value space to primitives, `SfItem<primitive>`, and arrays of those with a flat `params` record. Roughly 200ns in isolation, roughly 400ns end to end with the transform call, against a roughly 8.5µs report path, and only where a transform is configured. That let the isolation guarantee stand as written instead of being replaced with a caveat.

rfc/cmcd-reporter-middleware.md
littlespex: Fixed in f6dccc1, and the reproduction was worse than the finding states. The retry being suppressed is not the end of it: because target order is fixed by configuration, one consistently throwing transform starves every target after it for the rest of the session. First `update({ sta: 'p' })` threw, the retry did not even throw because dedup suppressed it before fan-out, and a later distinct `update({ sta: 'a' })` threw again with the healthy target still at zero reports.

Each target's transform is now invoked in isolation: a throw drops that target's report, the remaining targets are still processed, queues are still flushed, and the first error is re-thrown after the fan-out completes. The throwing target consumes no sequence number, as if it had returned `null`.

I chose isolation over transactional staging or dedup rollback because it keeps the propagation stance intact and leaves dedup semantics alone; the RFC already commits to cancellation not rolling back the baseline, and a throw should not behave differently. Regression tests cover the starvation case directly, including that the healthy target's `sn` stays contiguous.

rfc/cmcd-reporter-middleware.md
littlespex: Fixed in f6dccc1. The examples now carry their imports, and `scrubPii` is defined where the composition example uses it rather than referenced out of thin air.

This applied to more than the RFC: the user guide section I wrote on the implementation branch had the same defect, an implicitly typed `data` parameter and an undefined `scrubPii`. Fixed there too.

rfc/README.md
littlespex: Fixed in f6dccc1. Resolved in favor of the public-API rule: the exception list is now "bug fixes, internal changes with no public API surface, and documentation updates", followed by an explicit statement that the public-API rule takes precedence over size, since adopters live with a small public addition and we cannot take it back without a breaking change. "Small" is not an exception; "not public" is.

This RFC is itself the case in point, being a small public addition that satisfied both rules as written.

PR #391, opened 2026-07-22
fix(608): advance cursor on mid-row code (spacing attribute)

tobbee:
## Bug

A CTA-608 mid-row style code is a spacing attribute — per the standard it occupies one on-screen cell (a space carrying the new pen) and advances the cursor, so text following the code starts one column to the right. Cta608Channel.ccMIDROW only applied the pen via setPen and never advanced the cursor. As a result, colored/italic text positioned by an encoder that (correctly) accounts for the mid-row cell is rendered one column too far left. It's visible as a colored caption line drifting left of a white line on the same centered screen (white lines are unaffected because they use an indent PAC and need no mid-row code).

## Fix

Advance the write cursor by one column after applying the mid-row pen. Adds a regression test.

## Context

I found this when working on automatically generating colored wall-clock-time 608 captions in livesim2.
There will be a visible test vector from livesim2 relatively soon, but the regression test should be enough
for this project.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,3 +18,4 @@ Please add entries to the bottom of the list in the following format:
 - @dsilhavy [Daniel Silhavy, Fraunhofer FOKUS]
 - @N1Knight [Nicolás Caballero, Qualabs]
 - @valentinamgiusti [Valentina Giusti, Qualabs]
+- @tobbee [Torbjörn Einarsson, Eyevinn]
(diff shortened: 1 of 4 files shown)


littlespex (approved):

littlespex:
@tobbee Thanks for this. This fix is now available in `v1.0.3`:

https://github.com/streaming-video-technology-alliance/common-media-library/releases/tag/608-v1.0.3

https://www.npmjs.com/package/@svta/cml-608

Issue #327, opened 2026-02-22
feat(cmcd): improve CmcdReporter error handling and expose errors to consumers


littlespex:
## Summary

CmcdReporter's error handling needs improvement in two areas: aligning with CTA-5004-A (v2) recommendations for HTTP error responses, and providing a mechanism for consuming applications to observe and react to reporting errors.

## Current Behavior

- **410 Gone** — target is removed from the event targets map
- **429 / 5xx** — events are re-queued via a `.catch()` handler, but with no backoff, retry limit, or consumer notification
- **Other non-2xx (e.g. 400, 403, 404)** — silently treated as success; events are discarded
- **Network/requester errors** — re-queued with no limit, potentially growing the queue indefinitely
- **No consumer visibility** — all error handling is internal; the containing app has no way to know reporting is failing

### 1. Retry with backoff and limits (per spec)

The v2 spec recommends that for **429** responses, players implement a backoff strategy which "MAY include automatically increasing the number of messages aggregated per batch or increasing the delay interval between the transmissions of subsequent batches." For **5xx** errors, players "SHOULD appropriately handle [them] by batching and retrying with a back-off interval strategy."

Currently, failed sends are re-queued and retried immediately on the next processing cycle with no backoff. This can lead to hot loops against a struggling server.

- Add exponential backoff (or respect `Retry-After` header for 429)
- Add a configurable max retry count per batch; drop events after exhausting retries
- Consider a configurable max queue depth to bound memory usage

### 2. Buffer on connectivity loss (per spec)

The spec states: "In the event of lost connectivity with Event mode reports, a player SHOULD batch reports and attempt to deliver them once connectivity is restored."

Currently, network errors are re-queued but there is no distinction between connectivity loss and other failures, and no mechanism to detect restoration.

### 3. Handle additional HTTP status codes

- **400 Bad Request** — events are malformed and should not be retried; drop them and notify the consumer
- **403 Forbidden / 401 Unauthorized** — likely a configuration issue; surface to consumer rather than silently discarding

### 4. Error callback for consuming applications

Add an optional error handler to the configuration so apps can log, alert, or take corrective action:

```ts
type CmcdReporterConfig = {
  // ... existing config
  onError?: (error: CmcdReportError) => void
}
```

The callback should be invoked for:
- Non-retryable HTTP errors (4xx other than 429)
- Retries exhausted
- Target removal (410)
- Network/requester failures

### 5. Improve test coverage

[...]

PR #394, opened 2026-07-22
feat(cmcd): harden CmcdReporter custom key and custom event support

littlespex:
## Description

This PR ensures custom key/value fields (reverse-DNS hyphenated keys per CTA-5004-B) and custom events (`e=ce` with `cen`) can be sent reliably through `CmcdReporter`. An audit confirmed both mechanisms worked end to end but were untested at the reporter level, undocumented in the user guide, and carried one data-loss bug. The work landed as the reviewable slices below:

- **`test(cmcd)` — reporter-level coverage (WI-1, WI-2, WI-6).** 39 new tests locking in current behavior before changing any of it: custom keys through every injection point (`update()`, per-event `data`, `createRequestReport()`, `recordResponseReceived()` read-back and `data` param) in query, headers, and event modes; `enabledKeys` gating; `cen` emission, force-include, payload gating, and batching; and direct unit coverage for `prepareCmcdData`, the filtering choke point that previously had no dedicated test file.
- **`docs(cmcd)` — user-guide documentation (WI-3).** New "Custom Keys" section (naming rules including the runtime gates the `CmcdCustomKey` type cannot express, the explicit `enabledKeys` opt-in with no wildcard, value types and wire-format behavior, headers-mode placement) and a "Custom Events" section pairing `recordEvent(CUSTOM_EVENT)` with the target configuration required for delivery — the previous lone example invited a silent-drop misconfiguration. Cross-linked with the validation guide.
- **`feat(cmcd)` — expose `customHeaderMap` (WI-5).** The option existed on `CmcdEncodeOptions` but was never populated by the reporter, hard-wiring headers-mode custom keys to `CMCD-Request`. It is now available on the reporter config: mapped custom keys route to their configured shards, unmapped ones still default to `CMCD-Request`, standard keys keep their spec-defined shards, and the option is inert in query mode. API report regenerated (one optional property).

The `@svta/cml-cmcd` suite goes from 348 to 396 tests (0 failures); lint, `tsc --noEmit`, the full workspace test run, and the docs build are clean.

Refs: #327 (fixes the poison-batch/encode-failure slice of the error-handling story; backoff, retry limits, additional status codes, and an `onError` callback remain open)

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated

https://claude.ai/code/session_01GATF6dLZU8kHT2JoveJZJT

---
_Generated by [Claude Code](https://claude.ai/code/session_01GATF6dLZU8kHT2JoveJZJT)_

## Review Updates

- `isCmcdCustomKey` and `CmcdCustomKey` tightened to RFC 8941 serializable names (lowercase first letter, then `a-z 0-9 . -`, interior hyphen). The redundant key regex in `prepareCmcdData` is removed, and the validators now report non-serializable hyphenated keys as unknown keys.
- The serializability check that the first review comment asked to relocate no longer exists in any form: it moved to `@svta/cml-structured-field-values` as `isSerializableSfMember`, was then reworked into a tolerant-encode option, and was finally removed entirely with the WI-4 rollback. Nothing in this PR pre-validates or discards values.

- Fixed a latent whitespace bug found in the same `@svta/cml-structured-field-values` serializers: `whitespace` was applied through a default parameter, so passing any partial options object (e.g. `encodeSfDict(data, {})`) silently dropped the space RFC 8941 emits after each comma. It is now read from the property and defaults to on. Every in-repo caller already passes `whitespace: false` explicitly, so encoded output is unchanged repo-wide.


libs/cmcd/src/isSerializableValue.ts
littlespex, line 1: Wouldn't it make more sense for this to be in `@svta/cml-structured-field-values`?

libs/cmcd/src/CmcdCustomKey.ts
```diff
@@ -4,6 +4,14 @@
...
  * syntax when defining their own prefix.
  *
+ * This type is intentionally loose — it accepts any hyphenated string.
+ * At runtime a key must also satisfy {@link isCmcdCustomKey}
+ * (characters `a-z A-Z 0-9 . -`, with a hyphen that is neither the
+ * first nor the last character) and RFC 8941 key serialization
+ * (lowercase first character, then `a-z 0-9 - _ . *`) to reach the
+ * wire; keys failing either check are dropped during preparation.
+ * Lowercase reverse-DNS names satisfy both.
+ *
```
littlespex: Instead of leaving this loose, can it be tightened to only allow RFC 8941 compliant values? This would also include updating the `isCmcdCustomKey` helper to also only allow RFC 8941 compliant values.

libs/cmcd/src/isSerializableValue.ts
littlespex: Moved in 7b2d89d4e/f539c2a88: it's now `isSerializableSfMember` in `@svta/cml-structured-field-values`. Rather than porting the control-character regex, the helper runs the package's own serializers in a try/catch, so it covers every RFC 8941 failure mode (out-of-range integers/decimals, invalid token content, unsupported types, values nested in inner lists/params) and can't drift from serializer behavior.

Behavior change in cmcd: a typed-but-invalid custom value like `new SfToken('bad token')` is now stripped during preparation and the rest of the report survives, instead of reaching the encoder and voiding the payload. The `createRequestReport`/`sendEventReport` catch paths remain for values that only become invalid after token-field wrapping; their tests now exercise that seam.

libs/cmcd/src/CmcdCustomKey.ts
littlespex: Done in e269c558f. `isCmcdCustomKey` now requires a lowercase first letter, then `a-z 0-9 . -`, with an interior hyphen — the CMCD charset intersected with RFC 8941 key rules — and `CmcdCustomKey` is ``Lowercase<`${string}-${string}`>``, which rejects uppercase literals at compile time and still works as the index-signature key in `CmcdRequest`. Template literal types can't express the digit-leading/punctuation rules, so the runtime check stays authoritative.

This made the second key regex in `prepareCmcdData` redundant (removed): every key passing a filter is serializable by construction. Side effect: the validators now report uppercase hyphenated keys as unknown keys instead of treating them as custom.

libs/cmcd/src/isSerializableValue.ts
littlespex: Follow-up from review discussion: `isSerializableSfMember` is gone again. Pre-validating meant serializing every clean value twice per report, so the skip moved into the encoder itself — `SfEncodeOptions.skipUnserializable` makes `encodeSfDict`/`encodeSfList` omit members that fail RFC 8941 serialization within the single encode pass. cmcd enables it in `encodePreparedCmcd`/`toPreparedCmcdHeaders`, and `prepareCmcdData` no longer pre-validates values. Granularity also improves: a bad value now costs exactly its own key in every mode, including token fields that previously voided the whole request payload. Commits a1c9035e5/4a66ccdde.

littlespex: Correction to my previous note: `skipUnserializable` has been backed out of this PR (commits a1c9035e5/4a66ccdde removed; branch is back at f539c2a88). The tolerant-encode approach will be proposed separately.

This PR keeps the state that resolved this thread: `isSerializableSfMember` in `@svta/cml-structured-field-values`, with `prepareCmcdData` using it to strip unserializable values before encoding.

littlespex: Final outcome on this thread: `isSerializableValue.ts` is gone from the PR entirely, along with the sfv helper it became.

WI-4 has been rolled back (`8a2c86509`, reverting 1b68a546d/7b2d89d4e/f539c2a88). Both approaches — the pre-validate pass and the tolerant-encode option — kept unserializable data off the wire by silently discarding it, which gives the containing application no way to know a report was altered or why. That observability surface needs stakeholder input first, so the slice is deferred and tracked separately against #327.

Nothing in this PR pre-validates or discards values now; the throwing behavior on `main` is unchanged here. The custom-key tightening from `e269c558f` stays, and its reporter-level coverage is retained.

libs/structured-field-values/src/serialize/serializeDict.ts
```diff
@@ -48,13 +48,16 @@ import { serializeParams } from './serializeParams.ts'
...
+export function serializeDict(dict: Record<string, any> | Map<string, any>, options?: SfEncodeOptions): string {
 	if (typeof dict !== 'object' || dict == null) {
 		throw serializeError(dict, DICT)
 	}
 
 	const entries = dict instanceof Map ? dict.entries() : Object.entries(dict)
-	const optionalWhiteSpace = options?.whitespace ? ' ' : ''
+	// The RFC emits a single SP after each comma, so whitespace is on unless
+	// explicitly disabled. Checking the property (rather than defaulting the
+	// whole options object) keeps the default intact for partial options.
```
littlespex:
```suggestion
```

libs/structured-field-values/src/serialize/serializeList.ts
```diff
@@ -32,12 +32,15 @@ import { serializeItem } from './serializeItem.ts'
...
-export function serializeList(list: SfMember[], options: SfEncodeOptions = { whitespace: true }): string {
+export function serializeList(list: SfMember[], options?: SfEncodeOptions): string {
 	if (Array.isArray(list) === false) {
 		throw serializeError(list, LIST)
 	}
 
-	const optionalWhiteSpace = options?.whitespace ? ' ' : ''
+	// The RFC emits a single SP after each comma, so whitespace is on unless
+	// explicitly disabled. Checking the property (rather than defaulting the
+	// whole options object) keeps the default intact for partial options.
```
littlespex:
```suggestion
```

Issue #174, opened 2025-05-07
Add ID3 scheme ID URI


littlespex:
The scheme ID URI `https://aomedia.org/emsg/ID3` is used to identify ID3 content in emsg boxes. CML should expose this string as a constant in the ID3 namespace.

PR #398, opened 2026-07-24
[RFC] Child Reporters for CmcdReporter

littlespex:
## Description

RFC proposing a `createChildReporter()` method on `CmcdReporter`. It lets multiple player instances report into **one CMCD session**: one `sid`, one increasing `sn` sequence per event target, one stream of periodic reports, one `msd`. The main use case is an hls.js parent player and its interstitial asset players.

Full proposal: [rfc/cmcd-reporter-child-reporters.md](https://github.com/streaming-video-technology-alliance/common-media-library/blob/claude/cmcdreporter-spawn-clone-4gzey8/rfc/cmcd-reporter-child-reporters.md)

A child owns the state of its own playback: `cid`, `sta`, `pt`, `bl`, `nr`, its state-change dedup baseline, and its request-mode report config. All session state stays with the root reporter: the `sid`, the CMCD version, every `sn` counter, the batched event queues and timers, the session-wide `bg`, and the once-per-session `msd`. The child is a real `CmcdReporter`, so the same player code works with a root or a child. Session-owned config (`sid`, `eventTargets`, `version`) is rejected at compile time, and the reserved keys `sid` and `msd` are stamped by the reporter after all caller-supplied data, so no reporting call can override them either. A companion `activate()` method selects whose data goes into `TIME_INTERVAL` reports; it maps 1:1 onto hls.js's `INTERSTITIAL_ASSET_STARTED` / `INTERSTITIALS_PRIMARY_RESUMED`.

```ts
// The primary player owns the session
const reporter = new CmcdReporter({ cid: 'primary-movie', /* …, */ eventTargets: [/* … */] })
reporter.start()

// An interstitial asset player reports into the same session
const child = reporter.createChildReporter({ cid: 'ad-creative-42' })
child.update({ sta: CmcdPlayerState.PLAYING, pt: 0 }) // child-owned playback state
child.recordEvent(CMCD_EVENT_AD_START)                // goes into the root's queues with the next shared sn
child.createRequestReport(segmentRequest)             // shared request-mode sn, child's cid
child.activate()                                      // TIME_INTERVAL reports now use the child's data
```

### Motivation (short version)

With CMCD v2 tracking of interstitials, each `HlsAssetPlayer` wraps a full child `Hls` instance, and its `CMCDController` builds its own `CmcdReporter`. Even with a fixed `cmcd.sessionId`, every asset player restarts each target's Sequence Number at 0, re-sends `msd`, starts duplicate `TIME_INTERVAL` timers, and dedups state events on its own. Without a fixed `sessionId`, every asset even gets a different `sid`. All sequencing state is private to each reporter instance, and sharing one reporter instance between players corrupts per-playback state. Shaka's `InterstitialAdManager` has the same gap. dash.js (a single `MediaPlayer` across ad periods) shows that collectors already expect one `sid` to span content boundaries.

### Also in this proposal

- **Four prerequisite fixes to existing single-reporter behavior, landing in their own PR before this feature.** Three of them are what this design's guarantees rest on rather than merely adjacent: per-call `sid`/`msd` are merge-overridable (so "a child cannot change the session" is false without a fix), `msd` input is unvalidated (`0` rejected, negatives reach the wire, and first-write-wins would make a bad first value permanent), and `resetSession()` never clears `msd`/`msdSent` (children share that gate, so sharing a broken one spreads the bug to every player). The fourth: `update({ sid: undefined })` removes `sid` from all later reports.
- **A dependency on the shipped report transforms (#399).** `CmcdChildReporterConfig` derives its members from `CmcdRequestReportConfig`, so `transform` and `customHeaderMap` are child-configurable and inherited from the creating reporter, and future request-report config needs no edit to the child type.
- **`nr` guidance for non-rendered players**, with one invariant: the active reporter is the one that does not set `nr`.

This follows the `rfc/` process introduced in #390.

### Feedback wanted

The RFC's **Unresolved questions** section lists where community input would help most:

1. State-change dedup scope: per player (this proposal) or per session stream?
2. Validation of the active-reporter model for periodic reports with collector implementers, including `interstitialAppendInPlace` and PiP dual-presentation. The `nr` wiring settles with this.
3. Should request-mode `sn` be scoped per target? One shared counter means traffic to CDN A moves CDN B's sequence, which is hard to square with the spec's "increase independently per each combination of mode and target". This is today's single-reporter behavior, not something children introduce, but a primary CDN plus an ad CDN is the first common case where one counter describes two real targets.
4. Are `bsd` and the absolute counters (`bsa`/`bsda`/`dfa`) in scope? `bsd` is a once-per-mode-per-destination key, the same clause shape that justifies session-gating `msd`, but it has no gate at all.
5. `child.update({ sid })` and `child.update({ bg })` are silently ignored: is a dev-mode warning worth it?
6. `msd` first-write-wins changes root behavior: acceptable, or should the freeze apply only once a session has more than one reporter?
7. Locked-field ergonomics: discoverable `never`-typed locks, or a quieter `Omit`?
8. Child teardown flush scope: session-wide flush on `child.stop(true)`, or a no-op?

## Requirements Checklist

- [x] Unit Tests updated or fixed (N/A: docs-only RFC; the proposed test plan, including the coverage the prerequisite PR carries, is in the RFC's Testing section)
- [x] Docs/guides updated (N/A: the RFC is the document)
- [x] Changelog updated (N/A: RFC-only PR, no package changes, matching #390)


littlespex:
@nicolaslevy, @cotid-qualabs, @robwalch, @dsilhavy Here is a proposal that should help manage the complexities of maintaining consistent CMCD reporting across multiple players playing a single piece of content. The main use case is when a second player instance is needed to play interstitial or SGAI content in parallel with the main stream. Please let me know if you have any questions, comments or concerns.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Event-mode sequence numbers
+
+Every reporter in a session enqueues through the root's target map. `sn` is assigned at enqueue time (`target.sn++`) on the root's counters, the same place as today. JavaScript is single-threaded, so enqueue order is total, and the single owner dispatches batches in queue order. Delivery order across batches is not guaranteed under retry (today or with children): a batch re-queued after a 429/5xx can arrive after a later-`sn` batch that was already in flight. That is allowed by the spec, because `sn` is assigned at enqueue: `sn` orders the target's reports, not arrival time. The result per (sid, target) is one increasing sequence with no gaps, across the root and any number of children, with batches interleaving all players' events in call order. Events suppressed by dedup, or dropped for a missing required field, consume no `sn`, as today.
+
+### Request-mode sequence numbers
+
+`createRequestReport()` on any reporter takes the next number from the root's request-mode counter. `msd` on the request path is gated by the root's request-mode `msdSent` flag. The child's own `enabledKeys`/`transmissionMode` control the encoding.
```
littlespex: **[P1] Scope request `sn` by target.** A single request-mode counter is not independent per target. With requests to targets A, A, then B, B begins at `sn=2` and activity at A changes B's sequence. [CTA-5004-B](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#sequence-number) says sequence numbers increase independently for each mode-and-target combination. Define request-target identity—likely origin/CDN—and keep one shared counter per target across every reporter.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+	 * A child's `stop()` returns the active role to the root if that child
+	 * held it. An explicit `activate()` is always honored, including on a
+	 * previously stopped child; call it again after restarting a child that
+	 * is still presenting.
+	 */
+	activate(): void
+}
+```
+
+There is no requester parameter on `createChildReporter`: a child cannot change the session's transport. At runtime the implementation reads only `cid`, `enabledKeys`, and `transmissionMode` from the config. A session-owned field passed through `as any` or from plain JavaScript has no effect. This is a tested invariant, not an implementation accident.
```
littlespex: **[P1] Seal transient session fields.** Hardening child configuration and `update()` does not cover the `Partial<Cmcd>` arguments accepted by `recordEvent()`, `recordResponseReceived()`, and `createRequestReport()`. The current merge order allows `sid`, `v`, and `msd` supplied there to override stored state or bypass the `msd` gate. Define reserved-field precedence for every reporting path and test each one.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+| `bg` value + its dedup baseline | Session | The spec scopes `bg` to all players in a session. `update({ bg })` on any reporter writes through to root-owned session state, and `bg` is injected into every report next to `sid`/`v`, so an active child's periodic reports carry the session's backgrounded state. Its dedup baseline is session-level: root and child writing the same value emit one `b` event, fired from the writing reporter with that reporter's data. Page-level `visibilitychange` gives every player the same visibility, so last-write-wins is enough; aggregation for split-visibility cases is deferred (see Future possibilities). |
+| Request-mode `enabledKeys` + `transmissionMode` | Per player (inherited by default) | Request-mode encoding is per player/CDN. An ad CDN may need query params where the primary CDN accepts headers. |
+
+The session-scoped counters `bsa`/`bsda`/`dfa` ("since session initiation") and `bsd` (reported "once per reporting mode and report destination") get no cross-player aggregation in v1. They stay player-supplied numeric inputs via `update()`, because there is no obvious merge rule to impose (see Future possibilities).
```
littlespex: **[P1] Resolve session-wide metrics before v1.** Deferring `bsa`, `bsda`, `dfa`, and `bsd` aggregation permits contradictory or decreasing session totals under one `sid`. Either aggregate these values, provide an explicit session-level update API, or suppress child-local values when no correct session total is available.

littlespex (commented):
The root/session partition is promising, but I found three spec-level blockers and several integration gaps. In addition to the inline comments: (1) the hls.js companion work should add `nr` transitions for preloading/non-rendered asset players and the inactive primary player, and explicitly settle `interstitialAppendInPlace` behavior; (2) define valid `msd` input before making it first-write-wins—the current truthy/`isNaN` check rejects `0` but accepts infinities; and (3) PR #390 is still open and owns the proposed `rfc/README.md`, so merge ordering matters. The PR is docs-only and all current CI, CodeQL, and DCO checks pass.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+| Requester | Session | Only event-report POSTs use it, and only the root sends them. |
+| Data store (`cid`, `sta`, `pr`, `bl`, `mtp`, `pt`, `br`, `ot`, …) | Per player | These fields describe one playback. A preloading child's `sta: 'd'` must never overwrite the presenting player's data. |
+| `lastEmitted` dedup baseline | Per player | Dedup answers "did this player's state change". A shared baseline would suppress real transitions: an ad's `sta: 'p'` would hide the primary player's later identical resume. See Unresolved questions. |
+| `bg` value + its dedup baseline | Session | The spec scopes `bg` to all players in a session. `update({ bg })` on any reporter writes through to root-owned session state, and `bg` is injected into every report next to `sid`/`v`, so an active child's periodic reports carry the session's backgrounded state. Its dedup baseline is session-level: root and child writing the same value emit one `b` event, fired from the writing reporter with that reporter's data. Page-level `visibilitychange` gives every player the same visibility, so last-write-wins is enough; aggregation for split-visibility cases is deferred (see Future possibilities). |
```
littlespex: **[P2] Last-write-wins cannot implement `bg`.** The key means all players in the session are not visible; one reporter writing `true` cannot establish that condition, so last-write-wins can emit incorrect session state. [The specification](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#backgrounded) also says to omit it when visibility is unknown. Track visibility per reporter and aggregate it, or make `bg` root/application-owned.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+```
+
+There is no requester parameter on `createChildReporter`: a child cannot change the session's transport. At runtime the implementation reads only `cid`, `enabledKeys`, and `transmissionMode` from the config. A session-owned field passed through `as any` or from plain JavaScript has no effect. This is a tested invariant, not an implementation accident.
+
+### State partition
+
+| State | Scope | Rationale |
+|---|---|---|
+| `sid` | Session | The spec's session spans primary and interstitial content on one device. Children never store a `sid`. It is injected from the root when each report is created, together with the `sn`, so a root `sid` change needs no notifications. |
+| `v` / config `version` | Session | Stamped on every report. Mixed versions under one `sid` would confuse collectors. |
```
littlespex: **[P2] Preserve per-mode and per-target versions.** The one-version-per-session statement conflicts with existing per-event-target version configuration and the specification's [mixed-version guidance](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#player-processing-requirements). Request-mode version can be session-owned while each event target retains its root-owned version; do not promise every report under a `sid` uses one version.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+- **Child `flush()`/`stop(true)` flush the whole session's queues.** Per-child flushing would need per-item attribution in shared queues. Every asset teardown sends partly filled batches; the spec allows it, but batching efficiency drops for ad-heavy content.
+- **No inertness after teardown.** A destroyed but still referenced child that keeps recording (a stray async callback) adds events to the live session stream. v1 relies on garbage collection plus `stop()`; a `detach()`-style hard cut is deferred (see Future possibilities).
+- **Session-scoped counters without aggregation.** `bsa`/`bsda`/`dfa` and `bsd` remain per-player inputs in v1, so a multi-player session can emit per-player values for spec-session-scoped counters.
+- **Cross-bundle typing.** If a player bundles its own copy of `@svta/cml-cmcd` while the app injects a reporter built from another copy, runtime works (only methods are called), but TypeScript rejects the assignment because private members make the class nominal. Needs documented single-copy/dedupe guidance.
```
littlespex: **[P2] Use a structural injection type.** Requiring an injected `CmcdReporter` class makes the primary hls.js integration fail type-checking whenever the application and player resolve compatible but separate package copies. Dedupe documentation is fragile; type the injection boundary with a structural reporter interface while still returning genuine `CmcdReporter` instances from the factory.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,467 @@
...
+
+Related but not changed here: root `update({ v })` can still rewrite the version mid-session. Children are protected (`version` is locked and `v` comes from the session), but locking the root's `v` after construction is a behavior change outside this RFC's scope. It is flagged for reviewers.
+
+### Interaction with the report-transforms proposal
+
+This proposal composes with [Report Transforms for `CmcdReporter` (#390)](https://github.com/streaming-video-technology-alliance/common-media-library/pull/390) without changes to either. Event mode: a target's `transform` runs at enqueue on the root's targets, so it also sees child reports; `data.cid` identifies the source, and `sn` is assigned after the transform on the root's counters, exactly per that RFC's reporter-owned-fields rule. Request mode: the top-level `transform` belongs to the request-report config, so it would join `enabledKeys`/`transmissionMode` as playback-scoped child config, inherited from the creating reporter unless overridden.
```
littlespex: **[P2] Report transforms require a change.** The claim that #390 composes without changes is inaccurate: its request transform must be added to `CmcdChildReporterConfig`, inherited from the creating reporter, and handled at runtime. Make the dependency explicit or derive the child request configuration so the proposals cannot drift.

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed for `sid` and `msd`, corrected for `v`.

Verified against the built package:

- `recordEvent("ps", { sid: "INJECTED-SID" })` emits `sid="INJECTED-SID"`, carrying a shared `sn` from this session's counter.
- A per-call `msd` survives into the report after the gate has closed: three `ps` events with `msd` supplied on the third emit `msd` twice under one `sid`.
- Same on `createRequestReport()`.

`v` is already sealed, so it is dropped from the list. `prepareCmcdData()` re-stamps `v` from the encoding options on every report in both modes ([prepareCmcdData.ts#L176-L183](https://github.com/streaming-video-technology-alliance/common-media-library/blob/main/libs/cmcd/src/prepareCmcdData.ts#L176-L183)), so `recordEvent("ps", { v: 1 })` emits `v=2`. The same finding retires the RFC's own claim that root `update({ v })` can rewrite the version mid-session, which was also wrong.

Added a **Reserved keys are stamped, not merged** section. The mechanism is the one #399 already shipped: `queueTargetEvent()` stamps `e`, `sn`, and `msd` after a transform runs, so the fix extends that step to `sid` and makes the `msd` gate authoritative rather than additive. It lands in the prerequisite PR, since it changes root behavior too.

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed, and the RFC claim is withdrawn.

Both halves check out. Per-target versions are live today: a target configured `CMCD_V1` emits a line with no `v` key alongside a v2 target under the same `sid`. And the spec does not merely permit the mixture, it recommends it, so "mixed versions would confuse collectors" argued against the spec's own compatibility path.

New **Versions are per mode and per target** section states the narrower rule: version is root-owned, per mode and per target. Request path uses the top-level `version`, each event target uses its own, and `v` is re-stamped from the encoding options so a child emits whatever the mode and target it reports through are configured for. A child still sets neither, so the `never` lock stays.

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed, and the dependency is now on merged code: #399 shipped, so this is not a pending-proposal composition question.

The old text also contradicted itself inside one paragraph, claiming no changes were needed and then describing the change.

Rather than add `transform` by hand and risk the same drift, `CmcdChildReporterConfig` now derives its members from `CmcdRequestReportConfig`:

```ts
export type CmcdChildReporterConfig = Omit<CmcdRequestReportConfig, "version"> & {
	cid?: string;
	sid?: never;
	eventTargets?: never;
	version?: never;
}
```

That picks up `transform` and `customHeaderMap`, which the hand-written list had also missed, and anything added to request-report config later. Type-tested: all four request-mode members are accepted, all three locks reject on a literal, and `sid` is still rejected through a widened non-literal object, so the `never` locks keep the property that motivated them over a plain `Omit`.

rfc/cmcd-reporter-child-reporters.md
littlespex: Taking the fix, not the framing.

The core point holds: `bg` asserts something about every player in the session, so no single player can establish it, and last-write-wins can publish a false session-wide claim. Since the key's presence is itself the assertion, a false positive is the worst direction of error.

One correction: the spec says "This key SHOULD only be sent if it is TRUE," which is omit-when-false, not omit-when-unknown.

Chose root-owned over aggregation. `child.update({ bg })` is now ignored, like `child.update({ sid })`, and the value is still injected into every report so an active child's periodic reports carry it. Aggregation is the more literal reading and stays a future possibility, but it needs the root to hold a registry of live children in order to poll them, which is the reference this design avoids everywhere else: children are garbage the moment they are dropped, and a registry would turn every forgotten `stop()` into a leak. Visibility is a page-level fact the application observes once, so root-owned puts the key where the knowledge already is.

rfc/cmcd-reporter-child-reporters.md
littlespex: Valid, with one part pushed back on.

Not aggregating in the library. The reporter cannot know whether a given input is a session total or a per-playback delta, so summing would be wrong as often as right.

Taking the rest, and stating the consequence plainly instead of filing it under future work: a session whose players each supply local values emits *decreasing* "since session initiation" counts under one `sid`. New **Session-scoped counters are the application's job** section says the application owns `bsa`/`bsda`/`dfa` and must supply session totals.

`bsd` is the sharper version of this, and the RFC now says so. It is a once-per-mode-per-destination key, the same clause shape that justifies session-gating `msd`, yet it has no gate at all, for roots today or children here. Either it deserves the `msd` treatment or `msd`'s needs a narrower justification. Gating it changes existing root behavior, so it is an unresolved question rather than bundled in.

rfc/cmcd-reporter-child-reporters.md
littlespex: Valid reading of the spec, but misattributed, so this becomes an unresolved question rather than a change here.

The behavior is not introduced by children. `createRequestReport()` already increments a single counter regardless of host. Verified with no children involved:

```text
cdn-a.example → sn=0
cdn-a.example → sn=1
cdn-b.example → sn=2
cdn-b.example → sn=3
```

So under the "increase independently per each combination of mode and target" reading, this is already wrong today for a single reporter.

What this RFC changes is how much it matters, and the RFC now says that: one player usually talks to one CDN, so one counter approximated one target well enough, while a primary player plus an ad CDN is exactly where one counter starts describing two real targets. I withdrew the old Rationale claim that a single counter is "safe under any reading," which overstated it.

Not fixing it here, for three reasons now written into the question: origin is not target identity (one CDN fronts many hostnames, one hostname can front several CDNs), a per-origin map is unbounded where today the cost is one integer, and it changes existing root output, so it belongs with the other behavior changes rather than inside a feature addition.

rfc/cmcd-reporter-child-reporters.md
littlespex: Real problem, different fix.

Confirmed the failure with `tsc`: two declarations of a class with private members are not interchangeable, so a bare `CmcdReporter` annotation rejects a foreign-copy instance.

Not exporting a structural interface for it. `Pick` already drops private members, so the boundary is one type alias on the consuming side, verified assignable across copies:

```ts
type InjectedCmcdReporter = Pick<CmcdReporter, "update" | "recordEvent" | "createRequestReport" | /* … */>
```

The injection boundary in question is hls.js's own config field, so it can do this without a change here. A published `CmcdReporterLike` would have to track the class forever and would weaken the genuine-instance guarantee that makes `createChildReporter()`'s return type useful, against an in-repo rule not to add public exports without a clear use case. The recipe is now in the hls.js integration sketch and Drawbacks, and exporting the type stays a cheap additive follow-up if several players hand-roll the same alias.

littlespex:
Revised in e67682f, with `main` merged in. Each inline thread has a reply and is resolved. Answering the three items from the review body:

**1. `nr` transitions.** Taken, and it belongs here more than the review implies. A session with children is by definition a session where several players fetch media and at most one renders it, which is exactly what `nr` is for. It needs no new machinery, since `nr` lives in each reporter's own per-player store, but it does need wiring. New **Non-rendered players and `nr`** section states the invariant as one line, *the active reporter is the one that does not set `nr`*, and covers preloading asset players, the inactive primary player, and the `interstitialAppendInPlace` case. That last one now settles together with the active-reporter question rather than separately: whichever reporter is not active is the one marking its fetches non-rendered.

**2. `msd` input validation.** Taken, with one correction and one addition. `Infinity` is stored but the encoder drops it, so nothing invalid reaches the wire today. Negatives do: `update({ msd: -500 })` emits `msd=-500`. And `0` is silently rejected for being falsy.

The reason this blocks the design rather than sitting next to it: under today's last-write-wins these are ordinary input bugs that self-correct on the next write, but under first-write-wins a bad first value is permanent, and in the `Infinity` case it locks the session's `msd` to a value that can never be emitted at all, with no later valid value able to replace it. Rule is now finite and non-negative, which admits `0`. It lands in the prerequisite PR.

**3. Merge ordering.** Stale, in a useful direction. #390 is merged, and #399 shipped the implementation, so `rfc/README.md` is already on `main` and there is no conflict to sequence. This branch was 3 commits behind; `main` is merged in now. The consequence runs the other way: transforms are shipped code, so the child config's omission of `transform` and `customHeaderMap` was a real gap against `main` rather than a hypothetical one, which is why the config type now derives from `CmcdRequestReportConfig` instead of listing members.

---

Two of the RFC's own claims were wrong and are withdrawn, neither of them flagged in review:

- **Root `update({ v })` cannot rewrite the version mid-session.** The RFC flagged this for reviewers as a known gap outside its scope. `prepareCmcdData()` re-stamps `v` from the encoding options on every report in both modes, so neither `update({ v })` nor a per-call `v` changes the emitted version. Verified: both emit `v=2`. The `v`-locking sub-question is removed rather than answered.
- **A single request-mode `sn` counter is not "safe under any reading" of the spec.** It is not independent per target. See the `sn` thread.

**Scope change:** the two companion fixes became **four prerequisite fixes in their own PR, landing before this feature**, rather than `### Fixed` entries here. Three of them are not merely adjacent, they are what this design's guarantees rest on: reserved-key stamping (without it "a child cannot change the session's identity" is false), `msd` validation, and the `resetSession()` `msd`/`msdSent` reset (children share that gate, so sharing a broken one spreads the bug across every player). Each also changes existing single-reporter output, which deserves its own reviewable diff.

Unresolved questions went from 6 to 8. Request-mode `sn` scoping and the `bsd`/absolute-counter question are now tracked there rather than settled, so resolving those threads closes the review item, not the design question.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+export class CmcdReporter {
+	/**
+	 * Creates a child reporter that reports its own playback under this
+	 * reporter's session. The child owns its data store, its state-change
+	 * dedup baseline, and its request-mode encoding config. It shares the
+	 * session's `sid`, version, sequence numbers, queues, timers, `msd`, and
+	 * requester. Calling this on a child creates another child of the same
+	 * root (the hierarchy is flat).
+	 */
+	createChildReporter(config: CmcdChildReporterConfig = {}): CmcdReporter
```
littlespex: **[P1] Preserve the reporter's `customData` generic.**

`CmcdRequestReportConfig` and the returned `CmcdReporter` are both used bare here, so they default to `Record<string, unknown>`. Current main deliberately threads `C` through the reporter, transforms, and request arguments. A strict TypeScript reproduction showed that a child created from `CmcdReporter<{ kind: string }>` accepts `{ customData: { wrong: 1 } }` without a diagnostic. Please parameterize `CmcdChildReporterConfig` and preserve or safely constrain the parent's `C` in `createChildReporter()`.

littlespex (commented):
Reviewed the current RFC head. I found five blocking design issues and four additional correctness/API issues; details are inline.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+
+   **Resolution (2026-07-29): leave it shared, and document the reading.** `sn` keeps incrementing regardless of where segments are requested from, so an aggregator that sees all origins can reconstruct the entire playback session in order, even with ads delivered from different origins. Per-target counters cannot provide that cross-origin order. The spec's MUST clauses hold either way: each target observes a monotonically increasing sequence (a gapped subset of the session counter), and the counter resets to zero on a new `sid`; nothing requires gaplessness. "Increase independently per each combination of mode and target" is read as describing what collectors may assume (no coordination across targets), not as forbidding a shared source. The documented costs stand: a single origin sees gaps it must not read as loss, and traffic to one CDN advances another's numbers. Per-target counters remain the compatible refinement in Future possibilities if CTA WAVE settles what "target" means in request mode.
```
littlespex: **[P1] Scope request-mode `sn` per target.**

A global counter makes target B's sequence depend on reports sent only to A, so B cannot distinguish those gaps from missing reports. [CTA-5004-B](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#sequence-number) says sequences increase independently for each mode/target combination, and the [standards rationale](https://github.com/cta-wave/common-media-client-data/issues/132#issuecomment-2898157210) explicitly expects target-local values. Please use a configured target identity and a target-keyed counter instead of overloading `sn` for cross-origin total ordering.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+
+### Event-mode sequence numbers
+
+Every reporter in a session enqueues through the root's target map. `sn` is assigned at enqueue time (`target.sn++`) on the root's counters, the same place as today. JavaScript is single-threaded, so enqueue order is total, and the single owner dispatches batches in queue order. Delivery order across batches is not guaranteed under retry (today or with children): a batch re-queued after a 429/5xx can arrive after a later-`sn` batch that was already in flight. That is allowed by the spec, because `sn` is assigned at enqueue: `sn` orders the target's reports, not arrival time. The result per (sid, target) is one increasing sequence with no gaps, across the root and any number of children, with batches interleaving all players' events in call order. Events suppressed by dedup, or dropped for a missing required field, consume no `sn`, as today.
+
+### Request-mode sequence numbers
+
+`createRequestReport()` on any reporter takes the next number from the root's request-mode counter. `msd` on the request path is gated by the root's request-mode `msdSent` flag. The child's own request-report config controls the encoding.
```
littlespex: **[P1] Do not consume the shared request `msd` gate before the child's key filter.**

The current request path sets `requestTarget.msdSent = true` before `prepareCmcdData()` applies `enabledKeys`. Once request config becomes per-child, a preload child that excludes `msd` can report first, close the shared gate without emitting the key, and prevent a later root or child that enables `msd` from ever sending it. Please mark the gate sent only after the prepared output retains `msd`, and add mixed child/root key-set tests in both orders.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+| `update()` / `recordEvent()` / `recordResponseReceived()` | As today | Child-local store and baseline (`msd` writes through to the session; `sid` and `bg` are ignored); enqueues into the root's queues |
+| `createRequestReport()` | As today | Child's request-report config; shared session `sn`/`msd` |
+| `isRequestReportingEnabled()` | As today | Answers for the child's own `enabledKeys` |
+
+A child stays usable after `root.stop()`, exactly like the root itself: recording still enqueues, and batch-size sends still fire; only the timers are off. There is no session "destroy": releasing all references ends the session through garbage collection, as today.
+
+### Target disposal and failure re-queue
+
+Both stay in the root's send path, unchanged. An HTTP 410 removes the target from the single session map and silences it for every player at once. A 429/5xx puts the failed batch back at the front of the single per-target queue. Children add no second owner, so the failure modes of naive queue sharing (interleaved queue edits, repeated reprocessing, re-queueing into a removed target) cannot occur.
```
littlespex: **[P1] Scope HTTP 410 disposal to the current session.**

Today a 410 permanently deletes the target, while `resetSession()` only resets counters. A focused reproduction produced one request across `s1 -> 410 -> update({ sid: 's2' }) -> new event`; the `s2` event was never sent. A delayed `s1` response can likewise delete the target after `s2` begins. [CTA-5004-B](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#player-processing-requirements) limits suppression to the remainder of the current session. The design therefore needs target restoration on reset plus a session-epoch guard before applying asynchronous 410 responses.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+// Teardown needs nothing new: clearAssetPlayer() → child Hls destroy() →
+// CMCDController.destroy() → child.stop(true). Reporters released on the
+// error path (resetAssetPlayer) are handled the same way.
+```
+
+### Non-rendered players and `nr`
+
+A session with children is a session where several players fetch media and at most one renders it. That is exactly what `nr` ("non rendered") is for: "True when the content being retrieved by a player is not rendered as audio or video. The key SHOULD only be sent when it is TRUE."
+
+`nr` needs no new machinery, because it lives in each reporter's own data store, which is already per player. It does need wiring, and the companion integrations should do it. The invariant is one line: **the active reporter is the one that does not set `nr`.** Every other reporter in the session is fetching content nobody is watching.
```
littlespex: **[P1] Decouple `nr` from periodic-report activation.**

`activate()` chooses whose `t` reports are emitted; it does not establish that every other playback is unrendered. During PiP, overlays, or parallel presentation, both root and child may render, so this invariant sends false `nr=true` data for one. [`nr` is defined by the player's actual rendering state](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#non-rendered), and the [standards discussion](https://github.com/cta-wave/common-media-client-data/issues/174#issuecomment-3058040883) explicitly notes that all players can be active. Please determine `nr` per player independently of periodic ownership.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+
+- `sid`: `recordEvent('ps', { sid: 'other' })` puts `sid="other"` on the wire. On a child that would emit a foreign session ID into the root's queue, carrying a shared `sn` drawn from this session's counter.
+- `msd`: a per-call `msd` survives into the report whether or not the once-per-session gate has already fired, so it can be sent more than once under one `sid`, violating "MUST only be sent once per Session ID".
+
+So `sid` and `msd` must be stamped by the reporter **after** all caller-supplied data, in both modes. This is not a new mechanism. [#399](https://github.com/streaming-video-technology-alliance/common-media-library/pull/399) already established it for transforms: `queueTargetEvent()` re-stamps `e` and assigns `sn` and `msd` after a transform runs, so "a transform cannot bypass the target's `events` filter via `e` or break `sn` continuity", and the request path stamps its reporter-owned fields the same way. The fix extends that existing stamping step to cover `sid`, and makes the `msd` gate authoritative rather than merely additive. It lands in the [prerequisite PR](#prerequisite-fixes), because it changes root behavior too.
```
littlespex, line 324: **[P2] Explicitly seal root-owned `bg` after per-call data and transforms.**

The stamping algorithm names only `sid` and `msd`. After moving `bg` to root-owned session state, `child.recordEvent(..., { bg: true })`, `child.createRequestReport(..., { bg: true })`, or either transform can otherwise override a visible root and publish a false session-wide assertion. Please add `bg` to the post-transform stamp and cover every per-call/transform path; the proposed tests currently exercise only `child.update({ bg })`.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+```ts
+// config: one new field on the existing cmcd config surface.
+// Typed structurally, so an app that resolves a separate copy of
+// @svta/cml-cmcd can still inject. `Pick` drops the class's private
+// members, which are what make a bare `CmcdReporter` annotation
+// nominal and reject cross-copy instances. See Drawbacks.
+type InjectedCmcdReporter = Pick<CmcdReporter,
+	'update' | 'recordEvent' | 'recordResponseReceived' | 'createRequestReport' |
+	'createChildReporter' | 'activate' | 'start' | 'stop' | 'flush' |
+	'isRequestReportingEnabled'>;
```
littlespex: **[P2] This `Pick` remains nominal across package copies.**

The selected `createChildReporter()` method returns the full private-member-bearing `CmcdReporter`. A strict two-copy TypeScript reproduction still fails with `TS2322`: the factory return types have separate declarations of the private session property. `Pick` without the factory compiles. Please redeclare the factory with a recursively structural return type, export a structural reporter interface, or narrow this boundary so it omits the factory.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+
+The root owns the `TIME_INTERVAL` timers. A child's `start()` is a safe no-op, and its `stop()` never touches the timers, so creating and destroying asset players cannot duplicate or stop the periodic reports. By default, periodic reports use the root's data. When an interstitial starts presenting, make its reporter the active one:
+
+```ts
+child.activate()    // TIME_INTERVAL reports now use the ad's cid/sta/pt/bl
+// …ad ends…
+reporter.activate() // reports use primary data again
+```
+
+`activate()` is optional. Without it, periodic reports keep describing the root's playback. A child's `stop()` returns the active role to the root if that child held it, so normal teardown needs no extra code. Two details: the return-to-root happens on every `stop()`, not only at teardown, and `start()` never takes the active role, because that would let a preloading child take it from the presenting player. A controller that stops and restarts a child that is still presenting (for example, a runtime CMCD disable/enable toggle) must call `activate()` again after `start()`.
```
littlespex: **[P2] Nested teardown restores the wrong active reporter.**

The RFC supports creating a child from a child. For active child A -> active nested child B -> `B.stop()`, this rule activates the root instead of still-presenting A, so subsequent `TIME_INTERVAL` reports use the wrong playback. Please restore the previous/creating active reporter, or document and show the required explicit `A.activate()` in the nested SDK flow; normal teardown is not currently sufficient there.

rfc/cmcd-reporter-child-reporters.md
```diff
@@ -0,0 +1,583 @@
...
+
+First-write-wins only works if "accepted" is defined, and today it is not. The current guard is `if (data.msd && !isNaN(data.msd))`, which:
+
+- **rejects `0`**, a legitimate value, because it is falsy;
+- **accepts `Infinity`**, which is stored but then dropped by the encoder, so it never reaches the wire;
+- **accepts negatives**: `update({ msd: -500 })` emits `msd=-500` today.
+
+Under last-write-wins these are ordinary input bugs, self-correcting on the next write. Under first-write-wins they become permanent: an `Infinity` or a negative accepted first would lock the session's `msd` for its whole lifetime, and in the `Infinity` case lock it to a value that can never be emitted at all, with no later valid value able to replace it. So the validation is a precondition of this design, not a tidy-up alongside it. The rule becomes: accept a finite, non-negative number, which admits `0` and rejects the rest. It lands in the [prerequisite PR](#prerequisite-fixes) because it changes root behavior.
```
littlespex: **[P2] Reject or normalize fractional `msd`.**

"Finite, non-negative number" accepts `12.5`, and the current encoder emits `msd=12.5,v=2`, although [`msd` is integer milliseconds](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#media-start-delay). First-write-wins would then freeze that invalid value for the session and block later correction. Please require `Number.isInteger`, or normalize before accepting, and add a fractional test case.

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed with the repo's compiler against the built types: a child typed per the RFC's bare signature accepted `customData: { wrong: 1 }` with no diagnostic, while a `CmcdReporter<{ kind: string }>` parent rejected the same request (TS2353). The child inherits the creating reporter's `transform`, so it must inherit `C` with it.

v4 parameterizes both: `CmcdChildReporterConfig<C> = Omit<CmcdRequestReportConfig<C>, 'version'> & …` and `createChildReporter(config?: CmcdChildReporterConfig<C>): CmcdReporter<C>`, with an `@ts-expect-error` test pinning that a child rejects requests the parent's `C` rejects.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Taken, and the prior resolution is superseded. The linked comment itself is unreachable from this environment, but the issue that introduced `sn` (cta-wave/common-media-client-data#132) states the purpose in its own body: allowing collectors to identify missing or duplicate entries. A shared counter defeats that at every target, since each report to CDN A punches a gap in CDN B's sequence, so B cannot distinguish routine gaps from loss. The cross-origin total order the previous resolution defended is a purpose the spec never assigns to `sn`; collectors that need cross-target ordering have `ts`.

v4 design: `CmcdRequestReportConfig` gains an optional `target` identity string; the session keys its request-mode `sn` counters and `msd` gates by it; reporters sharing an identity share one increasing sequence; reporters that set none share the default identity, which preserves today's output byte for byte, so the change ships with the feature rather than the prerequisite PR. Identity is configured rather than inferred because origin is not target identity (one CDN fronts many hostnames, one hostname can front several CDNs), and the counter map stays bounded by the identities the player names.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed in both modes against the built package: with `msd` excluded from `enabledKeys`, the emitted report carries no `msd` yet the gate closes anyway (`requestTarget.msdSent === true` after a filtered request report; same per event target). Invisible today because a single reporter's key set never changes, so a gate spent without emission is a gate that could never emit; false once request config is per child, exactly as described.

v4 adds it as a prerequisite fix (six total now): the gate is consumed only when the prepared output retains `msd`, with mixed child/root key-set tests in both orders and both modes. With request-mode counters now keyed per target identity (see the `sn` thread), the request-path gates are per identity, mirroring event targets.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed, both variants, against the built package: `s1 -> 410 -> update({ sid: 's2' }) -> new event` produced one requester call total, with the `s2` event silently dropped and the target map empty; and a delayed 410 from the old session deleted the target after `s2` had begun. The spec scopes suppression to "the remainder of the current session".

v4 adds it as a prerequisite fix, since it is pre-existing single-reporter behavior: disposed targets are restored on session reset, and 410 handling checks a session epoch so a stale response from a previous session cannot dispose a live target. The Target-disposal section and the state-partition row now state the session scope, and the test plan covers both the restore and the delayed-response guard.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Taken. The machinery was already per player (`nr` lives in each reporter's own store, and the reporter never writes it automatically), but the one-line invariant overstated the wiring guidance: activation selects whose data feeds the `t` reports and establishes nothing about other playbacks' rendering.

v4 restates the rule: `nr` reflects each player's actual rendering state; in a fully sequential session that coincides with activation, but during dual presentation (PiP, overlays) the non-active player is still rendering and must not set it. The state-partition row now reads "the rendering player is the one that does not set it", and the item-2 resolution in Unresolved questions carries the same refinement. (The linked comment is unreachable from this environment, but issue #174's premise, multiple simultaneously active players, is exactly the dual-presentation case the restated rule handles.)

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed against the built package: per-call `bg: true` rides both paths today (`bg,e=as,...` in the event body; `CMCD=bg,...` on the request URL), and transforms run before the reporter-owned stamp, so both doors are real.

v4 extends the stamp: on every child path (`recordEvent()`, `recordResponseReceived()`, `createRequestReport()`, and both transforms) `bg` is stamped from session state after all caller-supplied data, exactly like `sid`. It stays in this feature rather than the prerequisite PR because root-owned `bg` is introduced here, so there is no today-bug to fix, and root per-call `bg` stays legal: for the root it is the owner writing its own key. The test plan now covers every child path, not just `child.update({ bg })`.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed with a two-copy reproduction on the repo's compiler: the `Pick` including `createChildReporter` fails TS2322 ("Types have separate declarations of a private property"), the factory-less `Pick` compiles, and a recursively structural alias compiles and accepts cross-copy instances.

Omitting the factory would break the ad-SDK flow (SDKs call `createChildReporter` on an injected reporter), so v4 takes the recursive redeclaration: `Pick<CmcdReporter, ...> & { createChildReporter(config?: CmcdChildReporterConfig): InjectedCmcdReporter }`. The hls.js sketch and the Drawbacks recipe are both updated, and the future `CmcdReporterLike` note now refers to the recursive shape.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed against the RFC's own rules: with child A active and grandchild B (flat, a child of the root) activated, `B.stop()` handed the role to the root while A's interstitial was still presenting.

v4 changes the rule rather than documenting around it: `stop()` returns the active role to the reporter that created the child. For direct children that is the root, so today's behavior is unchanged there; for the nested case it is the still-presenting A. Cost is one creator pointer per child, no registry, and references still point only rootward/creatorward, so dropped children remain plain garbage. A stopped or abandoned creator is the same stale-data hazard Drawbacks already documents, unchanged in kind. The nested restore is in the test plan.

---

rfc/cmcd-reporter-child-reporters.md
littlespex: Confirmed against the built package: `update({ msd: 12.5 })` emits `msd=12.5` on both paths (and `msd=-500` still emits too), while the spec types `msd` as integer milliseconds.

One amendment to the requested fix: v4 rounds rather than rejects. Players compute `msd` from float `performance.now()` deltas, and the reporter already rounds its derived timings in `recordResponseReceived()`, so `Number.isInteger` would silently reject the natural way to produce the value. The rule is now: finite, non-negative, rounded to the nearest integer. `0` accepted, `12.5` normalized to `13`, the rest rejected. The prerequisite-fix row and the validation test table include the fractional case.

---

PR #399, opened 2026-07-27
feat(cmcd): add per-placement report transforms to CmcdReporter

littlespex:
## Description

Implements the accepted RFC [`rfc/cmcd-reporter-middleware.md`](https://github.com/streaming-video-technology-alliance/common-media-library/blob/main/rfc/cmcd-reporter-middleware.md) (#390): an optional `transform` function on the request-report config and on each event-target config, letting integrators modify or cancel individual CMCD reports before they reach the wire.

```ts
	enabledKeys: ['br', 'bl', 'sid', 'cid', 'v', 'sn'],
	// Request mode: never decorate license requests with CMCD
	transform: (data, request) => request.url.includes('/license') ? null : data,
	eventTargets: [
		{
			url: 'https://collector.example.com/cmcd',
			events: [CmcdEventType.RESPONSE_RECEIVED],
			enabledKeys: ['url', 'rc', 'sid', 'v', 'e', 'ts', 'sn'],
			// Event mode, this target only: segment responses only
			transform: (data, request) =>
				request?.customData?.requestType === 'segment' ? data : null,
		},
	],
})
```

Placement determines scope, mirroring `enabledKeys`: the top-level `transform` governs `createRequestReport()` reports, and each event target's `transform` governs that target's event reports. A synchronous `(data, request) => Cmcd | null`; returning nullish cancels the report.

### Why

`CmcdReporter` had no interception point between assembling report data and handing it to the wire. The `requester` argument runs post-batching and post-encoding, so customizing data means decode-and-re-encode, cancelling one event inside a batch is not possible, and it never sees request-mode reports at all.

The concrete driver came from dash.js review: a reporting target can opt into `RESPONSE_RECEIVED` events for selected request types only, and two targets may deliberately share a collector URL while differing in events, keys, batch size, and response filters. Expressing that previously required one `CmcdReporter` per target, fanning every `update()` out to N reporters with N timers and duplicated session state. Per-target transforms let one reporter own all targets and keep player-specific taxonomy (request types) in the player.

### What a transform cannot do

The contract is bounded so a transform cannot emit an invalid or cross-contaminated report:

- `e`, `sn`, and `msd` are stamped after the transform runs. A transform cannot change a report's event type to bypass a target's `events` filter, break `sn` continuity, or replay the media-start-delay marker. A cancelled report consumes neither a sequence number nor `msd`.
- Keys CTA-5004-B requires for the event are restored if a transform removes them: `ts` on every event, the signalled field for state-change events, `cen` for `ce`, `ec` for the error event, `url` for `rr`. Restoration never fabricates a key that was already absent, since that is a call-site bug.
- Report data is copied per target, nested values included, so mutating an array or an `SfItem`'s `params` in place cannot reach the reporter's persistent data or another target's report.
- Transform exceptions propagate rather than being swallowed, but are isolated per target: the remaining targets still receive the report, queues are still processed, and the error surfaces afterwards.

### Post-acceptance amendments (RFC v4)

Reviewing the v3 implementation surfaced three contract-level defects, each reproduced against working code before and after the fix. The RFC was amended to v4 and merged before this PR; the last three bullets above are those fixes. Worth a reviewer's attention because they are the least obvious parts of the diff:

1. **Required keys.** v3 protected only `e`, `sn`, and `msd`, so `{ ...data, sta: undefined }` on a `ps` event emitted `cid="…",e=ps,sid="…",sn=0,ts=…,v=2`, which this library's own `validateCmcdEventReport()` rejects. Note `ts` was also exposed, not just the state fields.
2. **Nested copy.** The per-report spread is shallow and 16 CMCD keys are array-typed, so `data.ec.push(...)` in one target's transform corrupted the persistent array and surfaced in reports for targets with no transform at all. The copy preserves prototypes deliberately: `prepareCmcdData`, `CMCD_FORMATTER_MAP`, `validateCmcdValues`, and the encoder all branch on `instanceof SfItem`, and both a plain spread and `structuredClone` drop it, which would silently change wire output. `structuredClone` is rejected on correctness first, cost second (roughly 40% of a report against roughly 5% for the targeted copy, measured).
3. **Error isolation.** `emitEvent` commits the state-change dedup baseline before fanning out and never rolls it back, so an unqualified throw aborting the fan-out starved every target ordered after the throwing one for the rest of the session, with the caller's retry deduped away. Reproduced: first `update({ sta: 'p' })` threw, the retry did not even throw, and a later distinct `update({ sta: 'a' })` threw again with the healthy target still at zero reports.

### Notes for reviewers

- **Public API surface** grows by exactly two type aliases and two optional config members; see the `cml-cmcd.api.md` diff. No runtime exports added, so tree-shaking is unaffected, and the types erase at compile time.
- **Cost when unused** is one `undefined` check per report. The nested copy runs only where a transform is configured, so a target without one pays nothing even when a sibling target has one.
- **File layout deviates from the RFC** in one respect: it proposed a single `CmcdReportTransform.ts`, but every other file in the package is named after the type it exports, so this uses `CmcdRequestReportTransform.ts` and `CmcdEventReportTransform.ts`. Exported names and signatures are exactly as specified.
- **`plans/cmcd-report-transforms/steps.md`** is the execution plan, including a Task 6 recording the v4 amendments so it matches shipped behavior rather than the v3 contract it was written against.
- **Known follow-ups, deliberately out of scope.** `recordResponseReceived` pins `customData` to `{ cmcd?: Cmcd }`, a TypeScript weak type, so a player passing only its own keys (`{ requestType: 'segment' }`) needs a cast; the `PlayerRequest` alias in the transform tests exists solely to work around it, and widening that signature is a separate change. The RFC frontmatter is still `status: accepted` and should become `implemented` with `implemented-in: @svta/cml-cmcd@<version>` at release prep, once a version exists.

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated

Validation: `npm run lint` clean, `npm run typecheck` clean, `npm test -w libs/cmcd` 436 passing / 0 failing (2 pre-existing todos), of which 42 are new. `libs/cmcd/package.json` is untouched; the changelog entry sits under `## [Unreleased]`.


littlespex (commented):
Posting the three verified findings from the review: timer-start isolation, required-key enforcement, and request immutability.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -407,26 +508,81 @@ export class CmcdReporter {
...
+		// the persistent store and with the other targets' copies.
+		copyReportValues(item)
+
+		// Captured so a transform cannot strip a key the event requires.
+		// `CmcdKey` spans custom keys too, so index through a record view.
+		const requiredKey = CMCD_REQUIRED_EVENT_KEYS.get(type)
+		const requiredValue = requiredKey ? (item as Record<string, unknown>)[requiredKey] : undefined
+		const ts = item.ts
+
+		const report = transform(item, request)
```
littlespex, line 587: **[P1] Keep starting later interval targets after a transform throws**

`start()` calls `recordTargetEvent()` directly for the synchronous initial `TIME_INTERVAL` event. If this transform belongs to the first target and throws, `start()` exits before later targets receive their initial event or even have their intervals armed. I reproduced this at this head: the error propagated and the second healthy target's transform was never called (`healthyCalls === 0`). Retrying `start()` hits the same target again, so it remains starved. Please give initial timer setup the same continue-then-rethrow isolation as `emitEvent()`.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -407,26 +508,81 @@ export class CmcdReporter {
...
+			return
+		}
+
+		// Restore, never fabricate: a required key that was already absent
+		// before the transform ran was a caller bug, not a transform bug.
+		if (report.ts === undefined) {
+			report.ts = ts
+		}
+
+		if (requiredKey && requiredValue !== undefined && (report as Record<string, unknown>)[requiredKey] === undefined) {
```
littlespex: **[P1] Enforce required event keys after final filtering**

This restoration runs before `sendEventReport()` calls `encodeCmcd(..., options)`, where `enabledKeys` is reapplied. For an `ERROR` target whose allowlist omits `ec`, the identity transform produced `e=e,sid="s",sn=0,ts=...,v=2`; `validateCmcdEvents()` rejects it for missing `ec`. `rr` similarly loses `url`, and type-valid replacements such as `url: ""` are dropped during preparation because this check only handles `undefined`. Please enforce the invariant at or after final preparation so the wire payload—not just the queued object—retains valid required fields. The guide's suggestion to omit a required key from `enabledKeys` should also be corrected: state fields and `cen` are force-included, while `ec` and `url` currently make the event invalid.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -535,15 +698,36 @@ export class CmcdReporter {
...
+		let cmcdData: Cmcd | null = merged
+
+		if (transform) {
+			// The spread above is shallow, so nested values are still shared
+			// with the persistent store.
+			copyReportValues(merged)
+
+			// The caller's request is passed, not the internal clone, so a
+			// transform cannot alter the outgoing report through it.
+			cmcdData = transform(merged, request)
```
littlespex, line 757: **[P2] Prevent request mutations from leaking through the shallow clone**

The transform receives the original request, but the outgoing `report` only shallow-clones it. Nested `customData` values and object-valued bodies remain shared. I reproduced a transform that sets `request.customData.player.kind = "license"`; the returned report also contained `"license"`, and both nested objects were identical. Mutable `FormData` or `URLSearchParams` bodies have the same issue. This contradicts the documented guarantee that mutating the request cannot affect the outgoing report. Please isolate the request passed to the transform, or change the contract and type so mutation is not presented as ineffective.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in 4b47bcc. Reproduced first, and confirmed it is worse than a missed initial event: after `start()` threw, the healthy target had zero transform calls, and a retried `start()` threw on the same target again with the count still zero. Its interval was never armed, so it would have reported nothing for the entire session, not just at startup.

`start()` now wraps the synchronous initial event per target with the same continue-then-rethrow contract as `emitEvent()`. One ordering detail worth noting: the timer is armed *before* the initial event fires, so a throwing transform leaves the target in the state a successful `start()` would rather than silently disabling it. Disarming on throw would have been a silent drop, which the RFC's error stance rejects everywhere else.

Regression test asserts the healthy target receives its initial event and the error still reaches the caller.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in 4b47bcc, and splitting it in two, because the reproduction showed the bigger half was not transform-related.

An `ERROR` target whose `enabledKeys` omits `ec` emitted `e=e,sid="s",sn=0,ts=...,v=2` with **no transform configured at all**. `prepareCmcdData` force-included `e`, `ts`, `cen` and the state-change field after filtering, but `ec` and `url` were never added to that set, so the 2.4.0 mechanism had a two-key hole in it on main. Fixed where that mechanism lives, with its own changelog entry under Fixed, since it changes wire output for targets that never configure a transform. `cen` was already covered, so the gap was exactly those two keys.

The transform-specific half is the `url: ""` case you identified: restoration only triggered on `undefined`. The predicate is now `isValid` minus its `false` exclusion, plus an empty-array check. `false` has to stay usable because `bg: false` is real on a backgrounded-mode event and the encoder emits `?0` for it, so a naive `isValid` would have silently reverted a transform that legitimately cleared it. That inverse case has its own regression test alongside the empty string, empty list and non-finite `ts` cases, each asserting the result passes `validateCmcdEventReport()`.

You were also right about the guide. With force-inclusion in place the advice is wrong in the opposite direction now: omitting a required key from `enabledKeys` does not suppress it at all. Corrected to say required keys cannot be configured away, and leaving the event out of the target's `events` is the way to not send it. Same correction applied to the RFC in v5.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in 4b47bcc by changing the contract, which is the second option you offered.

Confirmed first: `report.customData.player.kind` came back `"license"` with `report.customData.player === player` identical. My existing test only mutated `request.url`, a top-level primitive that genuinely cannot leak through the spread, which is exactly why it passed while missing this.

The request parameter is now `CmcdTransformRequest`, a read-only view: all members `readonly`, `headers` a readonly record, and `customData` a `Readonly<Record<string, unknown>>`. Typing `customData` as an opaque record rather than `any` is the load-bearing part. I tried a recursive `DeepReadonly<HttpRequest>` first and it is theatre for this bug: `HttpRequest` defaults `customData` to `any`, and a deeply readonly mapped type over `any` is still `any`, so it rejects `request.url = ...` while silently permitting the nested write you reported. With `unknown` values, nested writes are rejected too. Verified all five vectors (url, method, headers, customData entry, nested value) and that a plain mutable `HttpRequest` is still assignable.

I did measure the isolate-instead option before choosing: a recursive readonly `Proxy` blocks everything at runtime without copying, but costs roughly 1670ns per report against a roughly 8.5us report path, about 20%, because players build a fresh request per media object so it cannot be cached across calls. That is four times the nested-data copy for a caller-owned object, so it is recorded as a rejected alternative rather than adopted. Deep-copying the request was rejected too, since player `customData` is arbitrary and a generic copy either drops prototypes or needs an unbounded walker.

Two limits are now stated rather than implied: a mutable `FormData`/`URLSearchParams` body has its own mutating methods that no type can block, and JavaScript callers get no enforcement. So the contract is that the request is immutable, not that the library makes mutation impossible. That is a genuine narrowing of what v3 promised, so it is called out explicitly in the v5 Final Decision for adopters who read the old wording as a guarantee, rather than being folded in quietly.

PR #402, opened 2026-07-28
feat(cmcd): make the transform request generic over customData

littlespex:
## Summary

- `CmcdTransformRequest` typed `customData` as `Readonly<Record<string, unknown>>`, so a transform reading the player's own fields needed bracket access under `noPropertyAccessFromIndexSignature`. #400 had already made `recordResponseReceived()` generic over `customData`, which left an asymmetry: the call site kept the player's type, but the transform handed that same request saw every value as `unknown`.
- Threads a type parameter through `CmcdTransformRequest`, `CmcdRequestReportTransform`, `CmcdEventReportTransform`, `CmcdRequestReportConfig`, `CmcdEventReportConfig`, `CmcdReporterConfig`, and `CmcdReporter`. Every parameter defaults to `Record<string, unknown>`, so existing code compiles unchanged and un-parameterized spellings keep today's behaviour.
- Inferred from the configuration rather than declared per transform. Annotating one standalone transform as `CmcdEventReportTransform<PlayerData>` types the `request` argument in every transform on that reporter, including the top-level `transform` slot; `new CmcdReporter<PlayerData>({ … })` states it explicitly. The cost to an adopter is one annotation per reporter.
- Amends `rfc/cmcd-reporter-middleware.md` to v6, since the transform signature it documents changes shape.

## Packages Changed

| @svta/cml-cmcd | 2.4.1 | feat |

No version bump. Change notes are under `## [Unreleased]`.

## API surface

The `cml-cmcd.api.md` diff is exactly the seven declarations gaining `<C = Record<string, unknown>>`, plus `customData?: Readonly<C>`. No rollup renames were triggered.

One thing worth a reviewer's eye: `recordResponseReceived<C>` now shadows the class's `C` in the rollup, and nothing in the generated surface signals that the two are unrelated. `tsc` and `eslint` both accept it.

## Deferred: constraining the call sites

The reporter's methods accept a request with any `customData`, so a reporter given a concrete `C` and handed a request that does not satisfy it compiles, and the transform reads `undefined` where its type promises a value. This affects `createRequestReport()` as well as `recordResponseReceived()`. The usual symptom is a silently cancelled report rather than an error.

Three fixes were built and measured against the real suite:

| | closes `rr` | closes `createRequestReport` | breaks existing callers |
|---|---|---|---|
| drop the method's param, use class `C` | yes | no | **yes** |
| `D extends C = C` + `R extends HttpRequest<C>` | yes | yes | **yes** |
| conditional constraint, inert on the default | yes | yes | no |

The break in the first two is the blocker: an adopter whose `customData` is declared with `interface` rather than `type` gets `Index signature for type 'string' is missing` on a **defaulted** reporter, because interfaces get no implicit index signature. That would partly undo what #400 shipped. The existing tests do not catch it because they all use type aliases.

The conditional form (`Record<string, unknown> extends C ? any : C`) was verified to close both holes while breaking nothing, and is deferred as a non-breaking follow-up because it only rejects code that is already wrong. It was held back because it puts `createRequestReport<R extends HttpRequest<CmcdReporterCustomData<C>> = HttpRequest<C>>` in front of every adopter reading the docs. Recorded in the RFC so the decision is not relitigated from scratch.

## Test Plan

- [x] Three new tests in `CmcdReporter.test.ts`: inference from a single annotation, the explicit `new CmcdReporter<PlayerData>` form, and a guard that the un-parameterized spelling still compiles with bracket access. Each was watched failing first, with the expected errors.
- [x] `npm run lint`, `npm run typecheck`, `npm run build -w libs/cmcd`, `npm test -w libs/cmcd` all exit 0. 450 tests, 448 pass, 0 fail, 2 pre-existing todo.
- [x] `npm run build -w docs` exits 0; the only cmcd warning is pre-existing in `validateCmcdEventReport.ts`, untouched here.
- [x] The new user-guide example was extracted and run standalone: it compiles, filters as documented, and the misspelt-literal claim in its comment was confirmed to be a `tsc` error.
- [x] `new CmcdReporter({ … })` with no type argument still compiles, covered by the ~60 pre-existing transform tests.


libs/cmcd/src/CmcdReporter.ts
```diff
@@ -270,16 +270,24 @@ type CmcdEventTarget = CmcdTarget & {
...
+ *
+ * @typeParam C - The shape of the player's `customData`. Defaults to
+ *                `Record<string, unknown>`.
+ *
  * @see {@link https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#reporting-modes-when-we-send-data | CTA-5004-B Reporting Modes}
  *
  * @public
  */
-export class CmcdReporter {
+export class CmcdReporter<C = Record<string, unknown>> {
```
littlespex, line 287: [P2] Constrain request inputs with the reporter's `C`

`C` only types the configured transforms; `createRequestReport()` still accepts an arbitrary `R extends HttpRequest`, while `recordResponseReceived()` introduces an unrelated method-level `C`. A `CmcdReporter<{ requestType: "segment" | "manifest" }>` accepts `{ customData: { requestTypo: "segment" } }` at both methods with zero TypeScript diagnostics, so the transform sees `requestType` as typed while receiving `undefined` and may silently cancel reporting. Please thread the class `C` into both request inputs, using the default-preserving conditional constraint described in the PR if necessary.

libs/cmcd/src/CmcdTransformRequest.ts
```diff
@@ -34,5 +48,5 @@ export type CmcdTransformRequest = Readonly<Omit<HttpRequest, 'customData' | 'he
 	 * Any custom data the caller attached to the request, including the
 	 * player's own fields.
 	 */
-	readonly customData?: Readonly<Record<string, unknown>>;
+	readonly customData?: Readonly<C>;
```
littlespex: [P2] Preserve readonly semantics for nested `customData`

`Readonly<C>` only makes the top-level properties readonly. For `C = { nested: { requestType: string } }`, `request.customData!.nested.requestType = "license"` compiles with zero TypeScript diagnostics. Because the reporter passes the caller-owned request directly to the transform, this can mutate player state despite the documented read-only contract. Please use a recursive readonly representation, or explicitly narrow the contract and documentation.

littlespex (commented):
Two verified type-safety findings.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in 24f571e0d.

Both methods now constrain the request against `C`. Your example is a compile error at both call sites, covered by a new test.

The constraint had to be inert on the default. Constraining directly against `Record<string, unknown>` rejects callers whose `customData` is declared with `interface` rather than `type`, since interfaces get no implicit index signature, which would partly undo what #400 shipped. The existing tests did not catch that because they all use type aliases. So it is the conditional form you pointed at, exported as `CmcdReporterCustomData<C>` because it appears in both public signatures.

Also renamed the new method type parameter to `RD`: both `D` and `T` collide in the api-extractor rollup and rename `CmcdRequestReport<D>` and `decodeCmcd<T>` as collateral. The api.md diff is now only the intended declarations.

The user guide and RFC described this gap as deliberately left open; both are updated in d181c2c7f.

libs/cmcd/src/CmcdTransformRequest.ts
littlespex: Fixed in 24f571e0d.

`customData` is now `DeepReadonly<C>`. Your example is a compile error, covered by a new test.

Worth recording why this was not just a shallow-vs-deep nitpick: RFC v5 argues that typing the values as `unknown` is what makes the read-only contract real, and states that nested writes are rejected. Supplying `C` reopened exactly that hole, so opting into typed reads silently cost you mutation safety. I verified the default path still rejects nested writes, so the regression was scoped to adopters who supply `C`, but the direction was wrong either way.

`DeepReadonly` went into `@svta/cml-utils` (8220332b9) rather than cmcd, since nothing about it is CMCD-specific. Functions are matched before the object branch so callbacks on `customData` stay callable, and it is a no-op over `Record<string, unknown>`, so the default is unchanged.

One consequence worth flagging: arrays inside `customData` are now `readonly T[]`. That is correct for the contract, but it is a tightening for anyone passing an array they mutate.

`createRequestReport` needed one internal cast, since `C` is assignable to `DeepReadonly<C>` for every concrete `C` but TypeScript cannot prove it while `C` is unresolved.

PR #416, opened 2026-08-11
fix(cmcd): enforce session-owned sid and msd and scope 410 disposal to the session

### LINKED ISSUE #408: CmcdReporter: per-call data and transforms can override reserved session keys (sid, msd)
Found while verifying the child-reporters RFC (#398). Affects `@svta/cml-cmcd` 2.5.0, both reporting modes.

## Summary

`CmcdReporter` stamps reporter-owned fields (`e`, `sn`, gated `msd`) after per-call data and transforms run, but `sid` is not stamped and `msd` supplied per call is not routed through the once-per-session gate. Callers and transforms can therefore emit reports under a foreign session ID or emit `msd` repeatedly, and an explicit `undefined` can remove `sid` from reports entirely.

## Reproduction

**Per-call `sid` override (event mode):**

```ts
	sid: 'real-session',
	eventTargets: [{ url: 'https://collector.example', events: ['ps'], enabledKeys: ['e', 'sid', 'sta', 'ts'] }],
}, requester)

reporter.recordEvent('ps', { sta: 'p', sid: 'INJECTED' })
// wire: sid="INJECTED" — per-call data spreads over the persistent store and is never re-stamped
```

The same override works in request mode via `createRequestReport(request, { sid: 'INJECTED' })`, and a target `transform` can rewrite `sid` the same way (only `e` and `sn` are stamped after the transform).

**Per-call `msd` bypasses the once-per-session gate (both modes):**

```ts
reporter.recordEvent(CMCD_EVENT_TIME_INTERVAL, { msd: 5000 })
reporter.recordEvent(CMCD_EVENT_TIME_INTERVAL, { msd: 5000 })
// msd reaches the wire on both reports; msdSent never flips
```

The gate in `queueTargetEvent()` / `createRequestReport()` only manages the value tracked via `update({ msd })`; a value riding the per-call data skips it, so `msd` can appear any number of times per session. CTA-5004-B defines `msd` as sent once per Session ID per active reporting mode.

**Explicit `undefined` poisons the store:**

```ts
reporter.update({ sid: undefined })
// the sid-change guard (`if (data.sid && ...)`) skips falsy values, but the merge
// `{ ...this.data, ...data }` copies the explicit undefined in.
// Every subsequent report now omits sid entirely.
```

## Expected

`sid` is reporter-owned session identity: reports always carry the reporter's current `sid`, and per-call data or a transform cannot substitute or remove it. `msd` only reaches the wire through the once-per-session gate.

## Proposed fix

Stamp `sid` (and route `msd` exclusively through the gate) after caller data and transforms in both modes, extending the reporter-owned stamping step introduced in #399 (`queueTargetEvent()` for event mode, the post-transform block in `createRequestReport()` for request mode). Report-time stamping from reporter-held session identity also fixes the `update({ sid: undefined })` poisoning for free.

### LINKED ISSUE #409: CmcdReporter: msd rejects 0, skips validation, and mishandles session reset and key filtering
Found while verifying the child-reporters RFC (#398). Affects `@svta/cml-cmcd` 2.5.0. Three related defects in the `msd` (Media Start Delay) lifecycle.

## 1. Validation rejects `0` and accepts invalid values

The guard in `update()` is:

```ts
if (data.msd && !isNaN(data.msd)) {
	this.msd = data.msd
}
```

- `update({ msd: 0 })` is rejected (falsy), but `0` is a valid instant-start value.
- `Infinity`, negatives, and fractions are accepted: `msd: 12.5` and `msd: -500` reach the wire as-is.

CTA-5004-B defines `msd` as integer milliseconds. Expected: accept finite, non-negative numbers and round to integer milliseconds (`0` accepted, `12.5` rounds to `13`, `Infinity`/`-1`/`NaN` rejected).

## 2. `resetSession()` never clears `msd` state

A `sid` change resets `sn` counters and the dedup baseline, but not `this.msd` or the per-target `msdSent` flags:

```ts
reporter.update({ msd: 800 })   // sent on the next report, msdSent = true
reporter.update({ sid: 's2' })  // new session
reporter.update({ msd: 950 })   // stored, but silently never sent — msdSent is still true
```

CTA-5004-B sends `msd` once per Session ID per active reporting mode, so a new `sid` must re-arm the gate.

## 3. The gate is consumed before the key filter

With `msd` absent from `enabledKeys`, the report omits `msd` but the gate is consumed anyway, so `msd` is never sent even after a config that includes it would apply. Verified in both modes:

- Request mode: `createRequestReport()` sets `msdSent = true` before `prepareCmcdData()` applies the `enabledKeys` filter.
- Event mode: `queueTargetEvent()` stamps `msd` and flips `msdSent` at enqueue time, but the filter runs later at encode time in `sendEventReport()`.

Expected: consume the gate only when the prepared output actually retains `msd`. This has no observable output change for a single reporter with a fixed config, but the flag must be truthful for session-state reasoning (and for the child-reporters work in #398).

### LINKED ISSUE #410: CmcdReporter: HTTP 410 target disposal outlives the session
Found while verifying the child-reporters RFC (#398). Affects `@svta/cml-cmcd` 2.5.0, event mode. Follow-up to the disposal cleanup in #360.

## Summary

CTA-5004-B scopes HTTP 410 suppression to the session: on 410 the client must cease reporting to that endpoint "for the remainder of the current session". `CmcdReporter.disposeEventTarget()` deletes the target from the `eventTargets` map permanently, so the suppression outlives the session, and the async response handling has no session guard, so a 410 from an old session can kill the new session's target.

## Reproduction

**Disposal survives a session change:**

```ts
const reporter = new CmcdReporter({ sid: 's1', eventTargets: [target] }, requesterReturning410)
reporter.recordEvent('e')          // requester called once, responds 410, target disposed
reporter.update({ sid: 's2' })     // new session
reporter.recordEvent('e')          // requester never called again — target map is empty
```

One requester call total; the new session reports nothing.

**Delayed 410 crosses the session boundary:**

A report sent during session `s1` whose 410 response resolves after `update({ sid: 's2' })` disposes the target inside session `s2`, which the spec does not permit (the 410 belongs to `s1`).

## Proposed fix

- Restore disposed targets on session reset (`sid` change re-arms every configured target).
- Guard the async 410 handling with a session epoch so a response from a previous session cannot dispose a current-session target.

littlespex:
## Description

Prerequisite fixes for the child-reporters RFC (#398). Verifying that RFC turned up bugs in how `CmcdReporter` handles session state. They land before the feature so its acceptance gate (a childless reporter keeps byte-identical output) means something. Two review rounds hardened the fixes into a structural change, per-session state retention, whose public design was accepted and merged as [`rfc/cmcd-session-retention.md`](https://github.com/streaming-video-technology-alliance/common-media-library/blob/main/rfc/cmcd-session-retention.md) (#417). This PR implements that contract as amended after acceptance (v8 through v13 in the RFC's revision history). The final shape differs from the accepted draft in one important way: session provenance is keyed by the `sid` itself, and the minted token is gone.

Fixes #408. Per-call data and transform return values could substitute `sid` or put `msd` on the wire outside the once-per-session gate, and `update({ sid: undefined })` removed `sid` from every later report. The reporter now owns `sid` as session state, stamps it into the pre-transform merge so transforms see the real session identity, and re-stamps it with `sn` and the gated `msd` after the transform in both reporting modes. A report's `msd` now comes only from the gate.

Fixes #409. `update({ msd })` rejected `0`, a valid instant-start value, and accepted `Infinity`, negatives, fractions, and values too large for an RFC 8941 structured-field integer. The guard now accepts finite values from `0` to `999_999_999_999_999` and rounds to integer milliseconds per CTA-5004-B. A `sid` change now clears the stored `msd` and re-arms every gate, since the spec sends `msd` once per Session ID. And the gate is consumed only when the prepared report keeps `msd`: both modes used to flip the flag before the `enabledKeys` filter ran.

Fixes #410. CTA-5004-B scopes HTTP 410 suppression to "the remainder of the current session" and to the target URL, but disposal deleted one target permanently. Disposal now silences every same-URL target in the session that sent the batch: a `sid` change starts fresh targets (re-arming intervals when the reporter is started), and a delayed 410 silences the session that sent the batch instead of the one that replaced it.

Two pin tests from the RFC's testing section land first and pass against the unmodified code: the 429 re-queue path (a re-queued batch is re-sent as-is after later-`sn` batches were delivered) and `sn` reset on `sid` change, which previously had no assertions.

## Session retention (implements the RFC as amended)

Review found six more session-correctness issues across two rounds: stale responses relabeled to the new session, re-queued batches crossing sessions, unencodable `msd` values closing the gate, single-config disposal when targets share a URL, partial batches destroyed by eviction, and archived snapshots mutable through caller-held arrays. Most were symptoms of session state being reset in place, so the outcome is structural: session-scoped state lives in per-session objects, retained per `rfc/cmcd-session-retention.md`.

- `sessionRetention` counts ended sessions kept in addition to the current one. Default `2`, `0` disables retention, `Infinity` never evicts. Only `number` values are accepted (floored); anything else falls back to the default instead of coercing.
- Every request returned by `createRequestReport()` carries a frozen provenance record on `customData` under the new `CMCD_REQUEST_PROVENANCE` export, a registry symbol. `CmcdRequestProvenance` is `{ sid, cid?, data? }`: the issuing session's `sid` (the attribution key), the `cid` in effect when the request was issued, and the request's per-call data as an encoded CMCD string. The record is stamped before the key filter and transform run, so it also rides requests the reporter does not decorate (transform-cancelled, request reporting disabled), and their responses report with full request-time data.
- `recordResponseReceived()` resolves the record's `sid` against the retained sessions, or drops the response. The resolved session supplies its frozen data snapshot, its next per-target sequence number, its `msd` gate, and its 410 disposal state. The record supplies the request-time layer: the decoded per-call data, and the issue-time `cid`, so a response that lands after a mid-session content change reports the content its request was about. The reporter never re-ingests the player-facing `customData.cmcd` object, so `SfToken` values keep their RFC 8941 wire type across JSON boundaries. Nothing is ever relabeled with the current `sid`, and a per-call `data.sid` is not an attribution key.
- Session identity is the `sid` itself. CTA-5004-B expects a `sid` to name one playback session, and the user guide now carries a warning that reusing one corrupts the integrity of the collected data: the reporter replaces the retained namesake, and the replaced session's late responses relabel onto the replacement. In exchange the record is constructible. Hand-built requests, and split topologies where one reporter decorates and another configured with the same session records, attribute by naming the `sid`. Spread and `Object.assign` carry the record through clones; `JSON.stringify` and structured clone drop symbol keys, and the exported symbol is the bridge (the record itself is JSON-safe and read by value).
- A `sid` change detaches the ended session's data graph, so the archived snapshot is frozen against later mutation of arrays the caller passed to `update()`, and drains the ended session's unsent reports before eviction can destroy them, so a partial batch survives even `sessionRetention: 0`.

## Notes for review

- Public API surface: `CmcdReporterConfig.sessionRetention`, the `CMCD_REQUEST_PROVENANCE` symbol, the `CmcdRequestProvenance` type, and the optional member on `CmcdRequestReport`. The `cml-cmcd.api.md` diff is exactly those.
- The accepted RFC evolved during implementation. The revision history (v8 through v13) records each amendment, ending at `sid`-keyed identity with the token retired, and the RFC body on this branch matches what the code ships.
- Wire output for single-session callers is unchanged; the record never reaches the wire in either transmission mode. The `session attribution` suite covers the cross-session behavior, with each review repro pinned as a regression.
- Judgment calls pinned by tests: an evicted session's stale response is dropped rather than relabeled, a reused `sid` replaces its retained namesake at the newest retention position (it never resumes it), archived sessions drain their remainders without waiting for `batchSize`, a record naming a never-seen `sid` drops, and the `RESPONSE_RECEIVED` request layer is the caller's untransformed per-call inputs, so a request transform's edits reach the request wire only.
- The design record is the RFC itself. The interim plan folder (`plans/cmcd-session-fixes/`) is removed as superseded; the deferred starvation-counter design stays in `plans/cmcd-session-counters/design.md` and waits for the child-reporters feature PR.
- The RFC stays `status: accepted` here. Release prep flips it to `implemented` with `implemented-in: @svta/cml-cmcd@<version>` once the version exists, per `rfc/README.md`.

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated


libs/cmcd/src/CmcdReporter.ts
```diff
@@ -614,28 +657,40 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
 	 * @param report - The finished report data.
 	 * @param type - The type of event being reported.
 	 */
-	private queueTargetEvent(target: CmcdEventTarget, report: Cmcd, type: CmcdEventType): void {
+	private queueTargetEvent(target: CmcdEventTarget, config: CmcdEventReportConfigNormalized<C>, report: Cmcd, type: CmcdEventType): void {
 		report.e = type
 		report.sn = target.sn++
-
-		if (!isNaN(this.msd) && !target.msdSent) {
+		report.sid = this.sid
```
littlespex: **[P1] Do not relabel stale response events.** A request created under session A can complete after a reset to session B. Restamping here emits B's `sid` with A's request URL and content data; I reproduced `cid="c1",sid="s2",sn=0` for a request created under `s1`. Detect the stored request `sid` mismatch and drop or separately attribute the stale response.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -858,6 +931,10 @@ export class CmcdReporter<C = Record<string, unknown>> {
 	 * @param data - The data to send in the event report.
 	 */
 	private async sendEventReport(config: CmcdEventReportConfigNormalized<C>, data: Cmcd[]): Promise<void> {
+		// Captured before the await: CTA-5004-B scopes 410 suppression to "the
+		// remainder of the current session", so a 410 that resolves after a
+		// session change belongs to a session that has already ended.
+		const epoch = this.sessionEpoch
```
littlespex: **[P1] Preserve the batch's originating session.** A failed session-A batch can be requeued after a sibling 410, restored in session B, and retried with B's epoch. A subsequent 410 then incorrectly disposes B's target and timer. Capture an origin generation per batch and prevent requeue or retry after that generation is invalidated.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -424,20 +448,34 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
 		}
 
-		if (data.msd && !isNaN(data.msd)) {
-			this.msd = data.msd
+		const { msd } = data
+
+		// CTA-5004-B defines msd as integer milliseconds, sent once per session.
+		// 0 is a valid instant-start value; anything non-finite or negative is
+		// ignored so it can neither reach the wire nor consume the send gate.
+		if (typeof msd === 'number' && Number.isFinite(msd) && msd >= 0) {
```
littlespex, line 596: **[P2] Reject unencodable `msd` values.** This guard accepts `1_000_000_000_000_000`, above the RFC 8941 Structured Fields integer maximum of `999_999_999_999_999`. Request mode then closes the gate before serialization throws, suppressing a later valid value; event mode retains a permanently failing batch. Validate the rounded value against the wire-format range before storing it. See https://www.rfc-editor.org/rfc/rfc8941#section-3.3.1.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -887,24 +966,49 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	 * `update({ sid })` restores the target.
 	 */
 	private disposeEventTarget(config: CmcdEventReportConfigNormalized<C>): void {
 		const target = this.eventTargets.get(config)
 		if (!target) {
 			return
 		}
 		this.disarmInterval(target)
-		this.eventTargets.delete(config)
+		target.disposed = true
```
littlespex: **[P2] Dispose all configs for the target URL.** This marks only one config entry. When multiple supported event-target configs share the same URL, the others continue sending after that endpoint returns 410. CTA-5004-B scopes suppression to the specific target URL, so disposal and restoration should cover every config with the matching URL. This behavior is pre-existing, but it is directly in scope for the #410 fix.

littlespex (commented):
Review findings for four session-correctness issues:

- stale response-received events can be relabeled as the new session;
- an old-session retry can acquire the new epoch and let a later 410 dispose the current target;
- finite but unencodable `msd` values close the gate before serialization fails;
- HTTP 410 suppression remains scoped to one config entry rather than the target URL.

I reproduced all four cases locally. The existing test suite and CI pass, so these paths need focused regression coverage.

libs/cmcd/src/CmcdReporter.ts
littlespex: Confirmed; reproduced the exact `cid="c1",sid="s2",sn=0` output. Before this PR the injection bug meant the stored `sid` accidentally won the merge, so the stale event went out as `sid="s1"` with the new session's `sn`; the stamping fix turned that into a clean relabel. Fixed structurally per the review discussion and the approved design in `plans/cmcd-session-fixes/session-state.md`: session state now lives in per-session objects retained under `sessionRetention` (default 3), and `recordResponseReceived()` resolves its session from the `sid` stored on the request, falling back to a per-call `data.sid` when request mode does not enable `sid`. The repro now emits `cid="c1",sid="s1"` with `s1`'s next `sn`, drawing snapshot keys from `s1`'s frozen data and, if still unsent, `s1`'s `msd`. A key naming no retained session drops the event rather than relabeling it. Landed in 5029df5a7 with the `session attribution` test suite.

libs/cmcd/src/CmcdReporter.ts
littlespex: Confirmed by trace: the re-queue closure had no staleness guard, so a batch re-queued after a sibling 410 survived the disposal's queue clear, was restored with the session, and its retry captured the new epoch. The origin-generation counter became unnecessary under the per-session restructure (5029df5a7): the closure holds the batch's own session's target, so a failed session-A batch re-queues into A, where A's `disposed` flag blocks any further drain and eviction ends it; the 410 path likewise disposes the batch's own session, so it can never touch the replacement session's target or timer. Pinned by `never retries a batch invalidated by its session disposal`.

libs/cmcd/src/CmcdReporter.ts
littlespex: Confirmed empirically: `1_000_000_000_000_000` passed the guard, `msdSent` flipped after preparation (which only checks finiteness), then `encodePreparedCmcd` threw `failed to serialize "1000000000000000" as Integer` and a later valid `msd` was suppressed; in event mode the queued batch failed on every retry. `update()` now validates the rounded value against `999_999_999_999_999` before storing it, so an over-range value is ignored like any other invalid input and the gate stays open. Test table extended with the maximum accepted, a value rounding down to the maximum accepted, and `1e15` rejected. Landed in 33f88f8f4.

libs/cmcd/src/CmcdReporter.ts
littlespex: Confirmed: disposal marked only the config whose batch received the 410, while same-URL sibling configs (a supported pattern since the per-target transforms) kept sending to the gone endpoint. `disposeEventTarget()` now sweeps every config in the affected session whose `url` matches (exact string compare), dropping their queues and disarming their timers when the session is the live one; restoration on `sid` change already covered all of them. Landed in d136d77f1, pinned by `disposes every same-URL config in the session on a 410`.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -682,6 +842,21 @@ export class CmcdReporter<C = Record<string, unknown>> {
 			return
 		}
 
+		// The event belongs to the session that issued the request. The sid
+		// stored on the request by createRequestReport() is reporter-written
+		// provenance and wins; a per-call sid lets the player attribute the
+		// response when request mode never stored one (sid not enabled).
+		const cmcd = request.customData?.cmcd ?? {}
+		const sid = cmcd.sid || data.sid
+		const session = sid ? this.sessions.get(sid) : this.session
```
littlespex: **[P1] Preserve request session provenance.** A supported request transform returning `null` leaves `customData.cmcd` empty. If that request completes after a `sid` change, this falls through to the current session and relabels the old response. String-only lookup also maps an original `s1` request onto a later fresh `s1` after `s1 → s2 → s1`. Preserve non-wire session-generation provenance on the returned request and drop the response if that generation is no longer retained.

littlespex (commented):
Follow-up review of the retained-session refactor at `f997576b1`.

The exact four repros from the first round are fixed and now covered by regression tests. The refactor still has two P1 data-integrity/data-loss paths, a snapshot-isolation bug, a required RFC-process gap for the new public option, and one normalization mismatch. I reproduced the runtime findings locally; the full repository test suite and current PR checks remain green, so these paths need focused regressions.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -424,32 +492,92 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+		this.sessions.delete(sid)
+		this.session = this.createSession(sid, { ...this.session.data })
+		this.sessions.set(sid, this.session)
+
+		// An evicted session's unsent queues die with it.
+		for (const key of this.sessions.keys()) {
+			if (this.sessions.size <= this.config.sessionRetention) {
+				break
+			}
+			this.sessions.delete(key)
```
littlespex, line 666: **[P1] Drain the ended session before eviction.** No processing pass runs when sessions are swapped. A sub-batch queue from `s1` therefore remains unsent after `update({ sid: 's2' })`; with `sessionRetention: 1` this loop immediately evicts it, and further SID-only transitions can eventually evict it at the default retention. Drain the ended session before eviction and add a direct partial-batch regression.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -424,32 +492,92 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	 *
+	 * The data store is carried over by shallow copy: `cid`, `br`, custom
+	 * keys and the rest survive a `sid` change as they always have, while the
+	 * archived session keeps its own frozen view. A reused `sid` replaces its
+	 * archived namesake: a `sid` change is a new session per CTA-5004-B,
+	 * never a resumption.
+	 */
+	private startSession(sid: string): void {
+		this.sessions.delete(sid)
+		this.session = this.createSession(sid, { ...this.session.data })
```
littlespex, line 644: **[P2] Actually freeze archived session data.** This spread copies only the top-level object. Arrays and `SfItem` values remain shared, so mutating a previously supplied `bl` array after the transition changes a stale `s1` response from `bl=(25000)` to `bl=(99000)`. This contradicts the documented frozen snapshot; detach the complete supported CMCD value graph at the session boundary.

libs/cmcd/src/CmcdReporterConfig.ts
```diff
@@ -32,4 +32,18 @@ export type CmcdReporterConfig<C = Record<string, unknown>> = CmcdRequestReportC
...
+	 * `sid` change report under the session that issued its request, with
+	 * that session's data snapshot and sequence numbers. `1` keeps only the
+	 * current session, so stale responses are dropped. `Infinity` never
+	 * evicts, which also retains every session's unsent report queues.
+	 * Values are floored; out-of-range and non-numeric values fall back to
+	 * the default.
+	 *
+	 * @defaultValue 3
+	 */
+	sessionRetention?: number;
```
littlespex, line 47: **[P2] Route this public option through an RFC.** `sessionRetention` is a new exported configuration field, but `rfc/README.md` requires every public API addition—even a small one—to go through the RFC process. Its design currently exists only under `plans/`. Move the public proposal into an RFC and obtain acceptance before shipping.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -236,11 +237,14 @@ function createCmcdReporterConfig<C>(config: Partial<CmcdReporterConfig<C>>): Cm
 		...rest
 	} = config
 
+	const sessionRetention = Math.floor(config.sessionRetention ?? NaN)
```
littlespex: **[P3] Validate instead of coercing retention.** The public contract says non-numeric values fall back to `3`, but `Math.floor` coerces `'1'` and `true` to `1` and throws for `Symbol` values. Check that the input is a number before flooring, while retaining the documented `Infinity` behavior.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in f171354f8. Requests now carry an opaque session token on customData under the new CMCD_REQUEST_PROVENANCE export (a registry symbol), stamped on every createRequestReport() path including transform-cancelled decoration and disabled request reporting. Sessions are keyed by token, so a reused sid retains both generations and a stale response resolves the exact session that issued it; an own-namespace token whose session was evicted drops instead of falling through to sid lookup. Regressions cover transform-cancelled attribution, s1 → s2 → s1 original-generation attribution and post-eviction drop, foreign and fabricated token fallback, and a JSON serialization bridge. Design per the accepted rfc/cmcd-session-retention.md.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in f171354f8. startSession() now runs a drain pass after archiving the ended session and before the eviction loop, so a partial batch leaves under its own sid and sequence numbers even at sessionRetention: 0, where eviction is synchronous inside update(). Regression: 'sends a partial batch from the ended session before eviction'.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in f171354f8. The ended session's data graph is detached at the transition with copyReportValues(), which is complete for the CMCD value space (arrays, SfItems and their params), so the archived snapshot is now the frozen view the docs promised. The new session starts from a shallow copy of the detached graph. Regression: mutating a previously passed bl array after the transition leaves the stale report at bl=(25000).

libs/cmcd/src/CmcdReporterConfig.ts
littlespex: Resolved: the proposal went through the RFC process as PR #417, was accepted, and is merged as rfc/cmcd-session-retention.md. The implementation in f171354f8 follows the accepted contract: sessionRetention counts ended sessions (default 2, 0 disables retention), and request provenance ships as the CMCD_REQUEST_PROVENANCE export.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in f171354f8. The input is type-checked before flooring: only number values are considered, so '1' and true fall back to the default instead of coercing and a Symbol no longer throws under ToNumber. Per the accepted RFC the accepted range is now >= 0 (0 disables retention) with default 2, Infinity retained. The normalization table covers '1', true, Symbol(), -1, NaN, undefined, 0, 2.5, and Infinity.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -333,34 +405,30 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-			}
-
-			// Armed before the initial event so a throwing transform leaves the
-			// timer in the same state a successful start() would, rather than
-			// silently disabling the target.
-			target.intervalId = setInterval(timeIntervalEvent, config.interval * 1000)
-
 			try {
-				timeIntervalEvent()
+				this.recordTargetEvent(this.session, target, config, CMCD_EVENT_TIME_INTERVAL)
```
littlespex: **[P1] Keep the `start()` fan-out on one session.** An earlier target transform can synchronously call `update({ sid: 's2' })`. This loop then retains the old target object but passes the new session, queueing an s2 report against s1 counters. The next normal s2 event reuses `sn=0`; the repro produced duplicate `sid=s2,sn=0` reports for one target. Capture the session once before fan-out and use it consistently.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -706,9 +914,36 @@ export class CmcdReporter<C = Record<string, unknown>> {
 			}
 		}
 
-		const cmcd = request.customData?.cmcd ?? {}
+		this.emitEvent(session, CMCD_EVENT_RESPONSE_RECEIVED, { ...cmcd, ...derived, ...data }, request)
```
littlespex: **[P1] Do not trust request CMCD as a frozen snapshot.** `request.customData.cmcd` overrides the retained snapshot but is neither detached nor necessarily valid after the documented JSON bridge. A request encoded with `tab=(3000)` emitted a late s1 response with `tab=(9000)` after caller mutation. JSON-round-tripping an `SfItem` also produces a plain object that makes encoding reject and requeue indefinitely. Store or rehydrate a canonical deep snapshot before merging it here.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -825,25 +1084,39 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-			if (queue.length < config.batchSize && !flush) {
-				return
-			}
+				const deleteCount = drain ? queue.length : config.batchSize
+				const events = queue.splice(0, deleteCount)
+				this.sendEventReport(session, config, events).catch(() => {
+					// Re-queue events that failed to send. The target belongs
+					// to the batch's own session, so a failure after a session
+					// change re-queues there, never into the current session.
+					target.queue.unshift(...events)
```
littlespex, line 1242: **[P2] Freeze queued reports before retry.** Archiving detaches only `session.data`; queued and in-flight reports still reference caller-owned arrays. A batch initially sent as `bl=(25000),sid=s1,sn=0`, followed by a session transition, caller mutation, and HTTP 429, retried as `bl=(99000)` with the same identity. Deep-copy reports when queued so retries remain byte-stable.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -706,9 +914,36 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	 * Resolves the session a response belongs to. Reporter-written provenance
+	 * is authoritative: a token from this reporter either names a retained
+	 * session or, when that session has been evicted, drops the response.
+	 * Falling through to `sid` lookup there would relabel a reused `sid`'s
+	 * retained namesake. A foreign or absent token falls back to the `sid`
+	 * chain, where the newest retained session wins a reused name, then to
+	 * the current session when no key exists at all.
+	 */
+	private resolveSession(token: unknown, sid: string | undefined): CmcdSession<C> | undefined {
+		if (typeof token === 'string' && token.startsWith(this.tokenPrefix)) {
```
littlespex: **[P2] Classify exact issued tokens.** Any string beginning with `tokenPrefix` is classified as reporter-owned. Appending data to a real token therefore drops instead of following the documented SID fallback; changing generation `:0` to predictable `:1` can attribute the original request to the next session. Track exact issued tokens, including eviction tombstones, rather than using prefix membership.

libs/cmcd/src/CmcdRequestReport.ts
```diff
@@ -9,6 +10,7 @@ import type { Cmcd } from './Cmcd.ts'
 export type CmcdRequestReport<D = unknown> = HttpRequest & {
 	customData: {
 		cmcd: Cmcd;
+		[CMCD_REQUEST_PROVENANCE]: string;
```
littlespex: **[P2] Preserve structural type compatibility.** Making the symbol member required breaks existing consumers that construct or mock the public `CmcdRequestReport` type. A previously valid `{ url, headers: {}, customData: { cmcd: {} } }` now fails with TS2322, despite the RFC classifying this as non-breaking. Keep the general member optional and refine `createRequestReport()`'s return type, or treat this as a breaking release.

libs/cmcd/src/CMCD_REQUEST_PROVENANCE.ts
```diff
@@ -0,0 +1,19 @@
...
+ * boundary and restore it verbatim afterward. Never fabricate or alter a
+ * token; a value the reporter did not write attributes by the `sid`
+ * fallback chain instead.
+ *
+ * Backed by the symbol registry so duplicated copies of this library in
+ * one bundle interoperate. The registry key is stable across versions.
+ *
+ * @public
+ */
+export const CMCD_REQUEST_PROVENANCE: unique symbol = Symbol.for('@svta/cml-cmcd/request-provenance')
```
littlespex: **[P2] Mark the registry lookup pure.** Rolldown treats this unannotated `Symbol.for()` as effectful and retains it, plus dependent reporter initialization, when a consumer imports only `CMCD_V2`. This contradicts the RFC's tree-shakeability claim. Mark the initializer with `/* @__PURE__ */` so unused provenance code can be removed.

littlespex (commented):
Review findings from the current PR head. The focused repros are described inline.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -706,9 +914,36 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+			return this.sessions.get(token)
+		}
+
+		if (!sid) {
+			return this.session
+		}
+
+		let found: CmcdSession<C> | undefined
+
+		this.sessions.forEach((session) => {
```
littlespex: **[P2] Index retained sessions by SID.** Every tokenless or foreign-token response scans all retained sessions to find the newest matching SID. With a large retention value or `Infinity`, this becomes O(n) per response and potentially quadratic over the session history, rather than the RFC's stated map lookup. Maintain a newest-retained-session index by SID and update it during insertion and eviction.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in d4736b899. start() captures the session before the fan-out and passes it consistently, so a transform that rotates the session mid-loop leaves the remaining initial events on the session that started, with its own counters; they drain as archived reports. Regression: 'keeps the start() fan-out on the session that started'.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in d4736b899, two-sided. The stored cmcd is detached at decoration time (copyReportValues over the prepared output), and recordResponseReceived() canonicalizes it before the merge: SfItem-shaped plain objects left by the JSON bridge rehydrate to SfItem when they carry params and unwrap to their bare value when they don't, so encoding re-wraps per key instead of throwing and re-queueing forever. One documented residual: a custom-key SfToken is indistinguishable from its string after JSON and degrades to one. Regressions: 'freezes the request-stored cmcd against caller array mutation' and 'encodes a JSON-bridged response report without poisoning the queue'.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in d4736b899. The nested-value copy in recordTargetEvent() is unconditional now (it previously ran only under transforms), so queued and re-queued reports are detached at queue time and a 429 retry re-sends the bytes the batch was queued with. Regression: 'retries a failed batch with the bytes it was queued with'.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in d4736b899. Classification is by exact issued tokens: retained sessions are keyed by token and eviction leaves a tombstone, so an altered or forged value was never issued, classifies as foreign, and follows the documented sid fallback; only exact issued tokens attribute. Tokens are opaque uuid() values and the prefix/generation scheme is gone. Regression: 'falls back to the sid chain for an altered provenance token'; the reused-sid eviction test pins the tombstone drop. RFC v9 records the change.

libs/cmcd/src/CmcdRequestReport.ts
littlespex: Fixed in d4736b899. The member is optional on CmcdRequestReport, so the type stays constructible; the reporter still stamps every request it returns, and the TSDoc says so. A compile-level pin ('keeps CmcdRequestReport constructible without the provenance member') fails typecheck if it regresses, and the RFC's type block is amended in v9.

libs/cmcd/src/CMCD_REQUEST_PROVENANCE.ts
littlespex: Fixed in d4736b899. The initializer carries /* @__PURE__ */ and the annotation survives into the dist bundle (verified in the built output). I could not reproduce the Rolldown measurement here since it is not part of this repo's toolchain, but the annotation is the mechanism it documents honoring.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in d4736b899. A newest-wins sessionsBySid index resolves the fallback in constant time; it is set on insertion and cleared on eviction only when the evicted session is the indexed holder of the name. Pin: 'resolves a tokenless sid to the newest retained namesake' (two retained generations plus an evicted one). RFC v9's performance section matches.

littlespex (commented):
Fresh review of the redesigned implementation at `6ab4e82a`. The focused repros are described inline. The package tests and repository typecheck are green, but these edge cases still produce data corruption, invalid output, lost reports, or API/runtime mismatches.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -706,9 +1004,31 @@ export class CmcdReporter<C = Record<string, unknown>> {
 			}
 		}
 
-		const cmcd = request.customData?.cmcd ?? {}
+		this.emitEvent(session, CMCD_EVENT_RESPONSE_RECEIVED, {
+			...(typeof cid === 'string' && cid ? { cid } : undefined),
```
littlespex, line 1008: **[P1] Preserve an absent request-time CID.** When provenance has no `cid`, this conditional spread contributes no override, so `recordTargetEvent()` falls back to the session's later `cid`. A request created before `cid` is known and completed after `update({ cid: 'c2' })` is incorrectly emitted with `cid="c2"`. Preserve the request-time absence, for example by overriding current session data with `cid: undefined`.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -733,6 +1053,38 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	private createRequestProvenance(data: Partial<Cmcd> | undefined, baseUrl: string | undefined): CmcdRequestProvenance {
+		const base = this.session.provenance
+
+		if (!data) {
+			return base
+		}
+
+		try {
+			const encoded = encodeCmcd(data as Cmcd, {
+				version: this.config.version,
```
littlespex, line 1076: **[P1] Keep provenance snapshots independent of request version.** Encoding the snapshot with the outgoing request version makes version 1 down-convert inner lists and discard v2-only keys. With reporter `version: 1` and a default-v2 response target, per-call `bl: [25000]` replays as invalid `bl=25000,v=2` instead of `bl=(25000)`. Store a canonical, lossless snapshot independent of request-wire settings.

libs/cmcd/src/decodeCmcd.ts
```diff
@@ -54,11 +59,13 @@ export function decodeCmcd(cmcd: string, options?: CmcdDecodeOptions): CmcdData
...
 
-	const sfDict = decodeSfDict(cmcd)
+	const sfDict = decodeSfDict(cmcd, options)
 
+	// Each dictionary member is reduced whole, not via `item.value`, so
+	// member-level and inner-list-level params survive decoding.
 	const result: Record<string, unknown> = {}
 	for (const [key, item] of Object.entries(sfDict as Record<string, SfItem>)) {
-		result[key] = reduceValue(item.value)
+		result[key] = reduceValue(item, options?.useSymbol)
```
littlespex, line 68: **[P1] Model the newly preserved decode representations in the return type.** The overloads still promise `Cmcd`/`CmcdData`, but `useSymbol` now returns `SfToken` or `symbol` for standard token fields, while parameterized inner lists return an outer `SfItem` rather than an array. TypeScript accepts operations such as `decoded.tab.map(...)` and `decoded.ot.toUpperCase()` that then throw at runtime. Expose return types that represent these shapes.

libs/cmcd/src/CmcdDecodeOptions.ts
```diff
@@ -14,4 +14,14 @@ export type CmcdDecodeOptions = {
...
+
+	/**
+	 * Controls how RFC 8941 token values are represented in the decoded
+	 * data. When omitted, tokens are reduced to plain strings, which cannot
+	 * be told apart from string values on re-encoding. `true` decodes
+	 * tokens as registry `Symbol`s and `false` as `SfToken` instances;
+	 * either preserved representation re-encodes as a bare token, so a
+	 * decode/encode round trip keeps the wire bytes.
+	 */
+	useSymbol?: boolean
```
littlespex, line 26: **[P2] Forward `useSymbol` through header decoding.** This option reaches the query and URL helpers, but `fromCmcdHeaders()` still calls `decodeCmcd(value)` without forwarding it. Consequently both `useSymbol: false` and `useSymbol: true` return plain strings for header values, contradicting the shared API and changelog. Forward the representation option while retaining the existing cross-shard `convertToLatest` pass.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -424,32 +567,118 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+		// An evicted session's unsent queues die with it, and its stale
+		// responses drop, because its sid no longer names a retained
+		// session. The current session was inserted last, so oldest-first
+		// eviction never reaches it, and `Infinity` retention never evicts.
+		for (const key of this.sessions.keys()) {
+			if (this.sessions.size <= this.config.sessionRetention + 1) {
+				break
+			}
+
+			this.sessions.delete(key)
```
littlespex, line 666: **[P2] Do not evict a session that an active fan-out is still producing.** With `sessionRetention: 0`, a target transform can rotate to `s2` while `start()` or `emitEvent()` is still producing reports for captured `s1`. This loop evicts `s1` before those reports are queued; they are then appended to an unreachable session and never sent. Defer eviction until active fan-outs finish or otherwise keep active sessions processable.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -614,30 +851,45 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+		if (!isNaN(session.msd) && !target.msdSent && config.enabledKeys?.includes('msd')) {
+			report.msd = session.msd
 			target.msdSent = true
 		}
+		else {
+			delete report.msd
+		}
 
-		target.queue.push(report)
+		target.queue.push(encodeCmcd(report, createEncodingOptions(CMCD_EVENT_MODE, config)))
```
littlespex, line 892: **[P2] Commit counters only after encoding succeeds.** `sn` and `msdSent` are advanced above before this call can throw. If serialization fails, no line is queued, but the next valid report uses `sn=1` and omits `msd`; request reporting has the same ordering. Stamp tentative values, encode successfully, and only then advance the sequence and gate.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -424,32 +567,118 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	 * the caller handed to `update()` can no longer mutate it, and the new
+	 * session starts from a shallow copy of the detached graph, so `cid`,
+	 * `br`, custom keys and the rest survive a `sid` change as they always
+	 * have. Sessions are keyed by `sid`, which CTA-5004-B expects to be
+	 * unique per playback session: reusing one replaces the retained
+	 * namesake, whose remaining state dies as if evicted and whose late
+	 * responses relabel onto the replacement.
+	 */
+	private startSession(sid: string): void {
+		this.session.data = copyReportValues({ ...this.session.data })
```
littlespex, line 643: **[P2] Fully detach mutable structured-field values.** The archival copy here is not deep for mutable bare-item objects inside `SfItem.value` or `params`. `SfToken` values and `Uint8Array` parameters remain shared, so later caller mutation can rewrite an archived response; a transform can likewise alter a sibling target's bytes. Clone those nested values as part of the supported CMCD graph.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -825,25 +1211,39 @@ export class CmcdReporter<C = Record<string, unknown>> {
 	private processEventTargets(flush: boolean = false): void {
 		let reprocess = false
 
-		this.eventTargets.forEach((target, config) => {
-			const { queue } = target
+		// Oldest session first, so an archived session's late reports leave
+		// ahead of current traffic.
+		this.sessions.forEach((session) => {
```
littlespex, line 1216: **[P2] Avoid scanning every retained session on each report.** This visits every retained session even when archived queues are empty. With `Infinity` retention, 1,000 transitions caused 501,500 callbacks and the next event scanned 1,001 sessions, making rotation cumulatively quadratic and ordinary reporting linear in history. Track only sessions with processable queues or process the affected session directly.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -333,34 +448,36 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
 		let failure: { error: unknown; } | undefined
 
-		this.eventTargets.forEach((target, config) => {
+		// The fan-out is pinned to the session that was current when start()
+		// was called: a transform can synchronously rotate the session, and
+		// pairing the new session with the old session's targets would let a
+		// report consume another session's sequence numbers.
+		const session = this.session
+
+		session.eventTargets.forEach((target, config) => {
```
littlespex, line 466: **[P2] Respect `stop()` during the `start()` fan-out.** If target A's synchronous transform calls `stop()`, it clears timers and sets `started = false`, but this loop continues and arms target B. Its timer then keeps reporting after `stop()`; the focused repro produced four later B reports. Recheck `started` before arming remaining targets and/or inside timer callbacks.

PR #417, opened 2026-08-12
docs(rfc): session retention in CmcdReporter

littlespex:
Draft proposal for retained per-session state: the sessionRetention config option, opaque session-provenance tokens on decorated requests (customData.cmcdSession), and attribute-or-drop resolution semantics in recordResponseReceived(). Converted from
plans/cmcd-session-fixes/session-state.md and corrected per the #416 follow-up review: provenance replaces sid-string-only attribution, drain before eviction, detached archived snapshots, and validation instead of coercion for sessionRetention.

Refs #416.

## Description

Please describe the changes:

- If it's a feature or enhancement please be as detailed as possible.
- If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.
- If this change includes work derived from another project, please:
  - Include links to the original project and the specific files that were used.
    - Provide the appropriate attribution in the `NOTICE` file, including the full license text if available.
    - Update "Thanks" section of the `README.md`.

## Requirements Checklist

- [ ] All commits have DCO sign-off from the author
- [ ] Unit Tests updated or fixed
- [ ] Docs/guides updated
- [ ] Changelog updated

### DIFF
--- a/rfc/cmcd-session-retention.md
+++ b/rfc/cmcd-session-retention.md
@@ -1,5 +1,5 @@
 ---
-status: draft
+status: accepted
 ---
 
 # RFC: Session retention in `CmcdReporter`
@@ -298,4 +298,8 @@ None. Earlier drafts left the default window and the foreign-token fallback open
 
 ## Final Decision
 
-Pending community review.
+**Decision:** Accepted as proposed in v7.
+
+**Rationale:** Review on [PR #417](https://github.com/streaming-video-technology-alliance/common-media-library/pull/417) shaped the proposal rather than contested it. `sessionRetention` counts ended sessions, with `0` disabling retention and a default window of two. Provenance moved from a string-named `customData` field to the exported `CMCD_REQUEST_PROVENANCE` registry symbol, with the serialization bridge as the escape hatch for boundaries that drop symbol keys. The foreign-token `sid` fallback (rule 3) is the shipped behavior. Every question raised during drafting was settled in-document (see Revision history), so the RFC was accepted with no unresolved questions.
+
+**Date:** 2026-08-12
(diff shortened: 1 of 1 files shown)


rfc/cmcd-session-retention.md
```diff
@@ -0,0 +1,266 @@
...
+ * that session's data snapshot and sequence numbers. `1` keeps only the
+ * current session, so stale responses are dropped. `Infinity` never
+ * evicts, which also retains every session's unsent report queues.
+ *
+ * @defaultValue 3
+ */
+sessionRetention?: number;
+```
+
+Normalization is validation, never coercion. Only `number` values are considered: a number is floored, and a floored value of `1` or greater is used, `Infinity` included. Every other input falls back to `3`, including numeric strings, booleans, `NaN`, `0`, negative numbers, and values like `Symbol` that would throw under arithmetic coercion.
```
littlespex: I think this should be the number of retained sessions, excluding the active session. Meaning `0` represents no retained sessions.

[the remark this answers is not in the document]
littlespex: Fixed in c0505a959. The paragraph now separates the two claims: the input is type-checked, never type-coerced (non-numbers fall back instead of converting), and a number is then normalized by flooring and a >= 0 range check.

littlespex: Changed in c0505a959. sessionRetention now counts ended sessions, excluding the active one, so 0 disables retention. The default is re-expressed as 2, which keeps v1's effective window (the current session plus two ended ones). Recorded as v2 in the revision history, and the default value stays listed under unresolved questions for community input.

littlespex: Amended in f16ebba5d: the default of 2 ended sessions is part of the proposal, removed from Unresolved questions (v3).

PR #421, opened 2026-08-15
docs: CmcdReporter architecture plan and session-retention RFC v14

littlespex:
## Summary

Plans an internal restructuring of `CmcdReporter` before the child-reporters work starts, and records one more rejected alternative in the session-retention RFC. Documentation only: the refactor itself lands later in its own PRs, task by task, if this plan survives discussion.

Revised after self-review (d5a48b9dc): the plan now also fixes the five open defects from the #416 final review instead of preserving them, and the ownership model covers the player, target, and request lifetimes the first draft deferred.

## Why restructure

The transforms RFC took six revisions and the session-retention RFC took thirteen. Nearly every amendment fixed an interaction between two features that nothing in the reporter's structure prevented: transforms and required keys, sessions and queues, provenance and cancelled transforms, sid reuse and retention. The class enforces its invariants through call-site ordering and comments, and the two queued features (child reporters, automatic starvation counters) each multiply the cross-products.

Two concrete problems drove the shape:

- `CmcdSession` holds `data` and `lastEmitted`, which the child-reporters RFC's own state partition table says are per player. Children would have to rework the object #416 just built and tested.
- Retention archives one session-wide data store, but children make data per player. The plan resolves this now: playbacks carry a stable pid, rotation archives per-playback snapshots, and attribution resolves `record.pid ?? root`, so today's records behave identically and the typed `pid` member arrives later via a child-reporters RFC amendment.

## What's here

- `plans/cmcd-reporter-architecture/architecture.md`: the target design. Four internal units with single ownership: a session ledger (with dirty tracking and eviction deferred past active fan-outs), per-playback state (data, provenance, dedup baseline, lazy epoch reset; `bg` is session-owned), pure pipeline stages (`applyReportPolicy`, `stampReport`) with counter commits after successful encoding, and per-target outboxes with callbacks bound at construction. `CmcdReporter` keeps its exact public surface as a thin facade.
- `plans/cmcd-reporter-architecture/steps.md`: six extraction tasks. Existing tests are never edited and every task gates on an empty `cml-cmcd.api.md` diff; the five #416 defect fixes each land with a red-then-green regression test and a changelog entry.
- `rfc/cmcd-session-retention.md` (v14): records a rejected alternative raised after #416 merged: an opaque request id resolved against a reporter-private map instead of the provenance record. Rejected chiefly on lifecycle grounds. Documentation amendment, no semantic change.

## Sequencing

Refactor before children: children need the playback/session split either way, and doing that surgery once is the point. Counters follow children, per `plans/cmcd-session-counters/design.md`.


littlespex (commented):
Self-review: these are blocking findings, but GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR. The plan is not executable as written, and the proposed ownership model still omits key player, target, and request lifetimes that motivated the refactor. I reviewed the reasoning in #390, #398, #399, #400, #402, #416, and #417. The v14 opaque-request-ID amendment itself looks sound; the blocking findings are in the architecture and implementation plan.

plans/cmcd-reporter-architecture/steps.md
```diff
@@ -0,0 +1,772 @@
...
+```
+
+---
+
+### Task 2: Split player state from session state
+
+The load-bearing task: `CmcdSession` stops holding the per-playback fields (`data`, `lastEmitted`), which move to a new `CmcdPlayerState` held by the reporter, per the child-reporters RFC's state partition. Archived sessions gain an explicit frozen `snapshot`. State-change acceptance becomes a function over player state.
+
+**Files:**
+- Create: `libs/cmcd/src/CmcdPlayerState.ts`
```
littlespex: **[P1] Rename this internal state unit.** `libs/cmcd/src/CmcdPlayerState.ts` and the `CmcdPlayerState` value/type already exist as public API for the CMCD `sta` enum and are re-exported from `index.ts`. Creating this file/type as written either overwrites that API or creates duplicate declarations, so Task 2 cannot pass the required empty API diff. Use a distinct internal name such as `CmcdReporterPlayerState` or `CmcdPlaybackState` throughout the plan.

plans/cmcd-reporter-architecture/architecture.md
```diff
@@ -0,0 +1,136 @@
...
+	eventTargets: Map<CmcdEventReportConfigNormalized<C>, CmcdEventTargetState>;
+	requestTarget: CmcdTargetStamps;
+}
+
+type CmcdEventTargetState = CmcdTargetStamps & {
+	outbox: CmcdOutbox;
+}
+```
+
+The `snapshot` member is the archived player data written at rotation, replacing today's arrangement where the archived session's live `data` field doubles as the snapshot. While a session is current, event assembly reads the player's live data; once archived, it reads the frozen snapshot. This is also the seam that answers the retention × children gap: with N players, rotation archives the issuing player's data per session (or, later, per session and player), and the provenance record can name the issuing player the same way it names the `sid`.
```
littlespex: **[P1] Resolve the player dimension now.** One session-level `provenance` and one `snapshot` still cannot attribute a late child response to the issuing player's persistent store. `sid` selects only the session, while `cid` is mutable and is not a stable player identity. The phrase "or, later, per session and player" defers the central retention × children redesign this plan is intended to avoid. Define stable player-aware provenance and per-player archived snapshots here, or put the required player snapshot on the request record.

plans/cmcd-reporter-architecture/architecture.md
```diff
@@ -0,0 +1,136 @@
...
+## The four units
+
+All new modules are internal: none is exported from `index.ts`, so the public surface and the API report are untouched.
+
+### 1. `CmcdPlayerState` (per player)
+
+What one playback owns, straight from the child RFC's partition table:
+
+```ts
+export type CmcdStateField = 'sta' | 'pr' | 'cid' | 'bg' | 'br'
```
littlespex: **[P1] Keep `bg` out of per-player state.** The child-reporter RFC makes the `bg` value and its dedup baseline root/session-owned, and this document says the same at line 119. Putting it in every player's `lastEmitted` creates independent baselines and leaves no canonical root value to stamp onto child reports. Move `bg` plus its baseline to session state. The player state also needs an observed session generation/epoch so every child's ordinary dedup baseline resets lazily after `sid` rotation.

plans/cmcd-reporter-architecture/architecture.md
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export type CmcdSessionState<C> = {
+	sid: string;
+	provenance: CmcdRequestProvenance;
+	msd: number;
+	/** Frozen data snapshot archived at rotation; undefined while current. */
+	snapshot?: Cmcd;
+	eventTargets: Map<CmcdEventReportConfigNormalized<C>, CmcdEventTargetState>;
+	requestTarget: CmcdTargetStamps;
```
littlespex: **[P1] Model request targets as a map, not one scalar.** Child-reporter RFC v4 resolved request-mode `sn` and `msd` state as session-owned per configured target identity, with a default identity preserving current output. A single `requestTarget` cements the pre-children shape and forces another session-state/pipeline redesign when children land. Make this keyed by request target identity now.

plans/cmcd-reporter-architecture/steps.md
```diff
@@ -0,0 +1,772 @@
...
+
+Extracted from `queueTargetEvent` (lines 873 to 891) and `createRequestReport` (lines 1155 to 1171); the msd-gate TSDoc from both sites moves onto the function:
+
+```ts
+export function stampReport<C>(report: Cmcd, session: CmcdSessionState<C>, stamps: CmcdTargetStamps, attachMsd: boolean, type?: CmcdEventType): boolean {
+	if (type !== undefined) {
+		report.e = type
+	}
+
+	report.sn = stamps.sn++
```
littlespex: **[P1] Do not commit sequence state before encoding succeeds.** `stamps.sn++` runs here, and the event caller marks `msdSent` before `encodeCmcd`. If any unrelated value is unserializable, no report is queued but the sequence number and gate are permanently consumed; #416's final review reproduced this exact corruption. Prepare candidate stamps, encode and enqueue successfully, then atomically commit `sn`, `msdSent`, and future `bsd` cursors.

plans/cmcd-reporter-architecture/steps.md
```diff
@@ -0,0 +1,772 @@
...
+- Constructor: `this.ledger = new CmcdSessionLedger(this.config.sessionRetention, createCmcdSessionState(this.config.sid, this.config.cid, this.config.eventTargets))`. Delete the private `createSession` method.
+- `startSession()` becomes:
+
+```ts
+private startSession(sid: string): void {
+	const snapshot = copyReportValues({ ...this.player.data })
+	this.player = createCmcdPlayerState({ ...snapshot })
+	this.ledger.rotate(createCmcdSessionState(sid, this.player.data.cid, this.config.eventTargets), snapshot)
+	this.processEventTargets()
+	this.ledger.evict()
```
littlespex: **[P1] Do not evict a session that an active fan-out still holds.** A synchronous transform can rotate from `s1` to `s2`; with `sessionRetention: 0`, this call evicts pinned `s1` before later targets enqueue into it. Those reports land on an object no longer reachable by the ledger and are never sent. Add a ledger lease/refcount or defer eviction until every active fan-out using the session finishes.

plans/cmcd-reporter-architecture/steps.md
```diff
@@ -0,0 +1,772 @@
...
+			sn: 0,
+			msdSent: false,
+		},
+	}
+}
+```
+
+- [ ] **Step 3: Rewire `CmcdReporter.ts`**
+
+- Both `createCmcdSessionState` call sites (constructor, `startSession`) pass `this.requester`. Leave the constructor parameter's and the `requester` field's declared types exactly as they are (the inline function type is assignable to `CmcdRequester` where the factory consumes it).
```
littlespex: **[P1] Initialize the requester before constructing initial outboxes.** The base constructor creates its initial session before assigning `this.requester`. Applying this instruction to both factory calls therefore uses `this.requester` before assignment under the repo's strict settings (and would otherwise give initial outboxes `undefined`). Assign the field first or pass the constructor parameter directly for the initial session.

plans/cmcd-reporter-architecture/architecture.md
```diff
@@ -0,0 +1,136 @@
...
+
+| Feature | Landing zone |
+|---|---|
+| Child reporters (root/child split, shared `sn`/`msd`/queues, per-player data and dedup) | Second facade over the shared ledger and outboxes; `CmcdPlayerState` instance per child; `activate()` pointer and root-owned `bg` in the ledger |
+| Automatic starvation counters (`bsa`, `bsda`, `bsd`) | Transition detection in `acceptStateChange`; totals, spans, and supplied-value flags on `CmcdSessionState`; per-target `bsd` cursors consumed in the stamp stage under the same filter-aware rule as `msd` |
+| Session retention (already implemented) | Absorbed: the ledger is the retention structure; archived snapshots become an explicit `snapshot` member instead of a repurposed live field |
+
+## Performance and bundle
+
+- No new per-report work. The deep copy still runs only where a transform is configured; assembly is the same spreads; stamping is the same field writes. The only new allocations are one `CmcdOutbox` per target at session creation (which replaces the plain queue-holding object created there today) and one `CmcdPlayerState` per reporter.
```
littlespex: **[P2] This zero-allocation claim does not match the steps.** `pendingStateEvents()` allocates an array on every `update()`, and outbox orchestration creates an `onGone` closure for every retained session/target on every processing pass, including empty and sub-batch outboxes. The plan also preserves the known linear scan across all retained sessions, so `sessionRetention: Infinity` makes ordinary reporting scale with history. Use allocation-free iteration, stable callbacks, and track only sessions with processable outboxes, or revise and measure the performance claim.

plans/cmcd-reporter-architecture/architecture.md
```diff
@@ -0,0 +1,136 @@
...
+- Module-level pure functions minify at least as well as private methods (no `this`, mangleable names, no prototype chain).
+- No new exports from `index.ts`, so tree-shaking for non-reporter consumers is untouched, and reporter consumers pull the same graph either way.
+- No side effects at module scope in any new file, per the repo rule.
+
+## Testing strategy
+
+The 4691-line `CmcdReporter.test.ts` suite drives the public API through a mock requester and encoded wire output. It is the specification. The refactor is complete only if that suite passes with zero edits at every step; any step that seems to require a test change is a behavior change and stops the work. No new public members means no new tests are required, and internal units are deliberately not tested directly (they are private implementation; the suite pins their composed behavior).
```
littlespex: **[P2] The unchanged suite is not the complete specification.** The exact #416 baseline was green while the pre-encode counter/gate corruption, active-fan-out eviction, incomplete structured-value copying, retained-session scan, and stop/start timer repros still existed. Adding characterization/regression tests is not a behavior change. Pin those cases before extraction, then explicitly decide which defects are preserved versus fixed.

plans/cmcd-reporter-architecture/steps.md
```diff
@@ -0,0 +1,772 @@
...
+- **Zero public API change**: no new exports from `libs/cmcd/src/index.ts`. After every task's build, `git diff --exit-code libs/cmcd/config/cml-cmcd.api.md` must pass.
+- All new files are internal modules under `libs/cmcd/src/`, named for their primary export, no default exports, `type` not `interface`, tab indentation, no module-scope side effects. Preserve every existing `/* @__PURE__ */` annotation when moving module-scope constants.
+- When moving code, move its TSDoc and comments with it verbatim. Do not add new rationale comments; rationale goes in the commit message.
+- Development requires Node >= 24. Fresh worktrees: `npm install` and `git submodule update --init` before anything else (without a local `node_modules`, typecheck resolves against another checkout's stale `dist`).
+- Tests import bundled output: always `npm run build -w libs/cmcd` before `npm test -w libs/cmcd`.
+- Every commit: `git commit -s`, conventional-commit format, trailer `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>`.
```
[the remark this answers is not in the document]
littlespex: Fixed in d5a48b9dc: the constraint now uses the AGENTS.md placeholder form, with the executing agent filling in its own name and model.

plans/cmcd-reporter-architecture/steps.md
littlespex: Renamed to `CmcdPlaybackState` throughout (d5a48b9dc): file, type, factory, and the facade field. Picked over `CmcdReporterPlayerState` for length and because it matches the child RFC's "state of its own playback" language. `CMCD_ROOT_PID` and `CmcdStateField` live in the same module, and Task 2 now names the collision explicitly so it does not get reintroduced.

plans/cmcd-reporter-architecture/architecture.md
littlespex: Resolved now rather than deferred (d5a48b9dc). Playbacks carry a stable `pid` (root = `CMCD_ROOT_PID`), the base provenance record moves to the playback (each child has its own `cid`, so per-playback records fall out naturally), rotation archives snapshots per pid on the outgoing session, and resolution is `snapshots.get(record.pid ?? CMCD_ROOT_PID)`. Every record minted today and every hand-built `{ sid }` record resolves to the root snapshot, bit-identical to current behavior. The typed `pid` member on `CmcdRequestProvenance` is public API, so it lands as a child-reporters RFC amendment; this refactor builds the structure and resolution rule but mints nothing. Snapshot semantics stay rotation-time with the record's issue-time `cid` override. Capturing the full store on the request record was considered and rejected in architecture.md: it changes the documented semantics and adds an encode to every request, including no-per-call-data requests that pay nothing today.

plans/cmcd-reporter-architecture/architecture.md
littlespex: Moved (d5a48b9dc): `bg` and `bgEmitted` are session state, `CmcdStateField` drops `'bg'`, `acceptStateChange` routes the bg entry to the session's value and baseline, `update()` strips `bg` from the playback merge the way it strips `sid` and `msd`, and assembly injects the session's `bg` into every report. The value carries across rotation while the baseline resets, matching today's store-carryover-plus-fresh-baseline behavior. Playback state gains `epoch` with a `syncPlayback()` check at the emit entry: inert for the eagerly-reset root, load-bearing for children's lazy reset.

plans/cmcd-reporter-architecture/architecture.md
littlespex: Now a map (d5a48b9dc): `requestTargets: Map<identity, CmcdTargetStamps>` with a single `CMCD_DEFAULT_REQUEST_TARGET` entry, so the wire is unchanged and the children PR adds the public `target` config key without reshaping session state again.

plans/cmcd-reporter-architecture/steps.md
littlespex: Fixed as part of Task 4 (d5a48b9dc): `stampReport` writes `stamps.sn` without incrementing, and commits happen only after the enqueue (event mode) or after prepare, encode, and attach (request mode), so a throwing encode consumes nothing. The task carries a red-then-green regression test: an oversized structured-field integer throws at enqueue, and the next report must be `sn=0` and still carry `msd`. Future `bsd` cursors are named at the same commit step for the counters plan.

plans/cmcd-reporter-architecture/steps.md
littlespex: Fixed as part of Task 3 (d5a48b9dc): the facade counts fan-out depth around both per-target loops (try/finally), `startSession`'s eviction goes through `requestEvict()`, which defers while the depth is nonzero, and `maybeEvict()` runs after the outermost fan-out and its processing pass, so the pinned session drains before it can be evicted. Regression test included: retention 0, a rotating transform on target A, asserting target B's s1 report is sent.

plans/cmcd-reporter-architecture/steps.md
littlespex: Fixed (d5a48b9dc): the constructor assigns `this.requester` immediately after config normalization, before any session construction; Task 5 shows the full constructor. Related change in the same revision: outbox callbacks (`onGone`, `onDirty`) are bound once per outbox at session creation, which moves event-target construction into the facade's `createSession`, so the factory now returns an empty target map.

plans/cmcd-reporter-architecture/architecture.md
littlespex: Revised (d5a48b9dc). `pendingStateEvents` is gone: `update()` iterates the exported `CMCD_STATE_FIELDS` table and fires `recordEvent` whenever the field is present, because the emit path's dedup already suppresses non-changes (the old pre-check duplicated that logic). Outbox `onGone`/`onDirty` are bound at construction. The ledger keeps a dirty set: enqueue and async re-queue mark, a pass takes the dirty sessions ordered by `seq`, so idle passes and `Infinity` retention cost set operations rather than scans. The performance section now enumerates the real allocations, including the one ordered array per pass that has work.

plans/cmcd-reporter-architecture/architecture.md
littlespex: Reframed (d5a48b9dc). The testing strategy now states the suite is necessary but not sufficient, existing tests are never edited, and enumerated additions are expected. Preserve-versus-fix is decided explicitly: all five findings you list are fixed, each in the task that restructures its code path, with a red-then-green regression test where the defect is wire-visible (copy depth in Task 1, mid-fan-out eviction in Task 3, counter/gate commits in Task 4, stop-during-start in Task 6) and the retained-session scan in Task 5 as performance-only. The changelog gains a Fixed section for the four wire-visible ones.

PR #422, opened 2026-08-15
refactor(cmcd): restructure CmcdReporter into ledger, playback, pipeline, and outbox units

littlespex:
## Summary

Restructures `CmcdReporter`'s internals into four units with single ownership, per the plan in [`plans/cmcd-reporter-architecture/`](https://github.com/streaming-video-technology-alliance/common-media-library/tree/main/plans/cmcd-reporter-architecture) (design discussion: #421): `CmcdPlaybackState` (per-playback data store, dedup baseline, base provenance record), `CmcdSessionState` + `CmcdSessionLedger` (per-sid counters and gates, retention, per-playback snapshots, dirty tracking), pure pipeline stages (`applyReportPolicy`, `stampReport`, with counters committed only after encoding succeeds), and `CmcdOutbox` (delivery only). `CmcdReporter` keeps its exact public surface as a thin facade.

Zero public API change: `index.ts` and `cml-cmcd.api.md` are byte-identical to main, and all ten new modules are internal. This is the structural groundwork the child-reporters RFC (#398) consumes: the playback/session split, identity-keyed request targets, session-owned `bg`, the rotation epoch, and per-playback archived snapshots all land here.

## Behavior changes

Five, all open findings from the #416 final review, each with a red-then-green regression test and a changelog entry. Everything else is bit-equivalent to main, verified per task against the built package.

1. A report value that fails structured-field encoding no longer consumes `sn` or the once-per-session `msd` gate.
2. Reports queued while a transform rotates the session mid-event are no longer lost to eager eviction at `sessionRetention: 0`, including on the interval-tick path.
3. Report copies now detach `SfToken`, `Uint8Array`, and `Date` internals, so caller or transform mutation cannot reach the persistent store, sibling targets, or archived snapshots.
4. `stop()` called from a transform during `start()` halts the remaining fan-out instead of leaving later targets' timers armed.
5. Event-report processing tracks only sessions with queued reports instead of scanning every retained session (performance only; `sessionRetention: Infinity` no longer scales ordinary reporting with session history).

## Reviewer notes

- The test suite is additions-only (325 added lines, 9 new regression tests) with exactly two modified lines: the private-state access paths at the old lines 1482 and 1513, repointed at the renamed internals. Both assertions and their meanings are unchanged.
- `update()`'s auto-fire pre-check is removed: the emit path's dedup performs the identical comparison against the same baseline, so suppression behavior is equivalent. This one is by analysis rather than a dedicated test.
- 16 commits in extraction/fix pairs per task; the repo squash-merges, so the split exists for review, not history.
- Full root `npm test` (lint, build all, typecheck, every package) passes; the api.md gate is clean for every package.
- A follow-up is already in flight for the one coverage gap the final review triaged as post-merge: a request-mode twin of behavior change 1's regression test (the event-mode half is pinned here).

## Review follow-up (2026-09-09)

The 2026-09-03 review's open threads are dispositioned. Six are fixed on this branch and resolved:

- The required-key restore baseline is captured before the transform copy (279a6a0a). Pre-existing on main. Changelog entry and two regression tests.
- `batchSize` is floored and clamped to at least 1 (6cd76b88). Pre-existing on main. Changelog entry and two regression tests.
- The nested-token detachment case has a test (cd24ba46).
- The unreachable `requestTargets` fallback is gone. One helper owns the stamp literal (7fa2f626).
- The ten new modules follow the comment conventions: TSDoc with the internal tag on every export, no new em dashes, no rationale comments, two drifted docs corrected (7c3430fe).
- The merge of main had misfiled this branch's unreleased changelog entries under the released 2.6.0 heading. They are back under Unreleased (53e3752b).

The four structural threads (fan-out helper, outbox dirty channel, prepared-output `msd` commit, copy dispatch collapse) are implemented in #452, stacked on this branch. They stay open until #452 lands.


littlespex (commented):
Review notes from a max-effort pass against main (merge-base 0dee24beb), with every claim below probed against both built bundles. Nothing here blocks the cmcd 2.6.0 release, which goes out from main first. This PR lands on its own afterwards.

What was checked. Main's full reporter suite against this branch's build passes except the two repointed private-state probes. This branch's suite against main's build fails 7 of the 10 new tests, and the other 3 pin fixes for bugs the branch introduced and then fixed. Wire output for `bg`, `msd`, rotation and late responses was compared across both builds. Bundle size was measured with rolldown.

Bundle. Importing only `CmcdReporter`, minified: 18,543 bytes on main, 21,349 bytes here (7,008 to 7,782 gzipped). Importing only `encodeCmcd` and `decodeCmcd`: 9,684 bytes on both, so tree-shaking for non-reporter consumers is intact.

The inline comments are ordered most severe first. Fixes for the tick eviction gap, the Buffer copy, the `bg` store handling, the changelog wording, and the plan document follow as separate commits on this branch, each with a reply on its thread. The rest are decisions or follow-ups.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -501,18 +231,38 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+		this.fanOutDepth++
+
+		try {
+			this.recordTargetEvent(session, target, config, CMCD_EVENT_TIME_INTERVAL)
+		}
+		finally {
+			this.fanOutDepth--
+		}
+
+		this.processEventTargets()
```
littlespex: `processEventTargets()` and `maybeEvict()` run after the `try/finally`, so a transform that throws on an armed tick skips both. If that transform also rotated the session at `sessionRetention: 0`, `evictPending` stays set and the ended session stays in the ledger until some later fan-out completes at depth 0. `stop()` and `flush()` never run `maybeEvict()`.

Probe against both builds: after such a tick this branch's ledger holds `[s1, s2]`. A `recordResponseReceived()` for a request issued under s1 is then sent as `e=rr,sid="s1",sn=1`. On main it is dropped, because `startSession()` evicted synchronously. `start()` and `emitEvent()` are not affected, since they catch per target before their post-steps.

Fix: run the two calls in the `finally`, or mirror `emitEvent()` (catch, post-steps, rethrow). Regression test with mock timers: the tick throws, then the late response must not be sent.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -603,22 +355,30 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-		session.data = { ...session.data, ...data, sid: undefined, msd: undefined }
+		this.playback.data = { ...this.playback.data, ...data, sid: undefined, msd: undefined }
+
+		// bg is session-owned too, but its value is a report field rather than a
+		// stamp: it is tracked here and picked up when each report is built, and
+		// the key is deleted from the store rather than left undefined. Dedup and
+		// emission stay in the event path.
+		if ('bg' in data) {
+			session.bg = data.bg
+			delete this.playback.data.bg
```
littlespex: `update({ bg: undefined })` used to leave a `bg` key with an undefined value in the store. In request mode with `enabledKeys: ['bg']`, that key made `prepareCmcdData` count a non-empty key set and force-add `v`, so main emitted `?CMCD=v%3D2` where this branch emits a bare URL. That is an improvement, but it is a wire change outside the five listed ones and has no changelog line. Every other key still behaves the old way after an explicit-undefined clear, so `bg` is now the one inconsistent key.

The new test "does not leave a bg key in the store when update() never received one" passes on main too, so it does not pin this change. A sibling test that does `update({ bg: true })` then `update({ bg: undefined })` and asserts the bare URL would.

Same line, separate cost: `delete` on the store object switches it to V8 dictionary mode until the next `bg`-free `update()` rebuilds it (`bg` is never the last key added, `sid` and `msd` follow it). Measured on the built bundle, `createRequestReport()` in that state allocates about 500 bytes more and runs 10 to 20 percent slower. Main never deleted from the store. Stripping `bg` from `data` before the merge avoids the delete.

libs/cmcd/src/applyReportPolicy.ts
```diff
@@ -0,0 +1,91 @@
...
+	// The caller's merge is shallow, so nested values are still shared with
+	// the persistent store, and in event mode with sibling targets' inputs;
+	// a transform gets its own detached copy so in-place mutation reaches
+	// neither.
+	copyReportValues(report)
+
+	const ts = report.ts
+	// Captured so a transform cannot strip a key the event requires.
+	// `CmcdKey` spans custom keys too, so index through a record view.
+	const requiredValue = requiredKey ? (report as Record<string, unknown>)[requiredKey] : undefined
```
littlespex, line 68: Copilot's comment on line 88 is right, and the behavior is pre-existing: main has the same sequence at lines 826 to 851. `requiredValue` points at the array the transform receives, so in-place mutation changes the value the restore step compares and restores.

Probe, identical on both builds. Transform `d => { d.br.length = 0; return d }` on a `bc` event with `br: [1000, 2000]` sends `br=()`, and `validateCmcdEventReport` accepts it. `d.br.length = 0; delete d.br` sends no `br` at all. `d.br.pop(); delete d.br` restores the mutated `br=(1000)`. Reassignment (`{ ...d, br: [] }`) is restored correctly, so only in-place mutation escapes. `br` and `ec` are the only array-typed required keys.

Zero-cost fix: capture `requiredValue` before `copyReportValues(report)` on line 63. `copyReportValues` never mutates its input, and the report is encoded synchronously and not retained, so handing the store's own array back is safe. The TSDoc above also says empty lists are dropped downstream. They are not: `encodeCmcd({ br: [], e: 'bc', ... })` emits `br=()`.

Left out of the fix commits because it is outside the plan's ledger of behavior changes. It needs a decision: ride here with a changelog line, or a separate PR.

libs/cmcd/src/copyReportValues.ts
```diff
@@ -0,0 +1,102 @@
+import type { Cmcd } from './Cmcd.ts'
+
+/**
+ * Copies a mutable object held inside an item: an SfToken-shaped value,
+ * a byte sequence, or an inner list of either.
+ */
+function copyBareValue(value: unknown): unknown {
+	if (value instanceof Uint8Array) {
+		return value.slice()
```
littlespex: `value.slice()` copies a plain `Uint8Array`, but on a Node `Buffer` (a `Uint8Array` subclass) `slice()` returns a view over the same memory. A byte sequence supplied as a `Buffer` is therefore not detached at any of the three copy sites.

Probe against this branch's build: two event targets, `update({ 'com.example-b': new SfItem(Buffer.from([1, 2, 3])) })`, target 1's transform sets byte 0 to 9. Target 2's POST and every later report carry `:CQID:` instead of `:AQID:`. The same write-through happens at archival and through the request's `customData.cmcd`. With `new Uint8Array([1, 2, 3])` everything is detached as intended. Exposure is Node only (server-side relays, tests).

Fix: `new Uint8Array(value)` here and on line 41, verified on a patched bundle. The detachment test should get a `Buffer` variant.

libs/cmcd/CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ and this project adheres to
...
 - Corrected the `CmcdEncodeOptions.version` documentation, which still claimed a default of `1`. Encoding has defaulted to `CMCD_V2` since 2.0.0: when `version` is not set, the version is inferred from the data's `v` key, falling back to `CMCD_V2`, which is why `encodeCmcd({ br: 1000 })` emits `br=1000,v=2`. Documentation-only change; runtime behavior is unchanged
+- A report value that fails structured-field encoding no longer consumes a sequence number or the session's once-per-target `msd` gate
+- Reports queued while a transform rotates the session mid-event are no longer lost when the outgoing session is evicted (`sessionRetention: 0`)
+- `SfToken`, `Uint8Array`, and `Date` values are now detached when reports are copied, so caller or transform mutation cannot alter the persistent store, sibling targets, or archived session snapshots
```
littlespex: This line describes two different pre-PR failures as one. Nested values (`SfToken` or `Uint8Array` inside an `SfItem`'s `params` or `value`) were shared on main, and "detached" is the right word for them. Bare `Date` and bare `Uint8Array` custom values were not shared on main. They were destroyed: the copy produced a prototype-only object with no internal slot, so every `createRequestReport()`, both transform paths, and the first report after a `sid` change threw (`TypeError: this is not a Date object.` and `Error: failed to serialize ... as Byte Sequence`). That crash fix is a sixth behavior change the plan's ledger excludes.

Also, a bare top-level `SfToken` was already detached on main (the new test at test line 4734 passes on main), and the nested-token case that is new has no test.

Line 30 understates the eviction fix as well: at `sessionRetention: 1`, main lost both of an archived session's reports, not only at 0.

Suggested: keep this line for the sharing cases with qualified wording, add a separate Fixed line for the crash, add a nested-token test, and record the sixth change in `plans/cmcd-reporter-architecture/steps.md`.

libs/cmcd/CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ and this project adheres to
...
+- Calling `stop()` from a transform during `start()` no longer leaves the remaining targets' interval timers armed
+
+### Changed
+
+- Internal restructuring of `CmcdReporter` into session-ledger, playback-state, report-pipeline, and outbox units in preparation for child reporters and automatic session counters. No public API change. Event-report processing now tracks only sessions with queued reports instead of scanning every retained session
```
littlespex: Bundle impact is not mentioned anywhere in the PR. Measured with rolldown (minify, ESM) against the built dist:

| import | main | this branch |
|---|---|---|
| `CmcdReporter` only | 18,543 B (7,008 gz) | 21,349 B (7,782 gz) |
| `encodeCmcd` + `decodeCmcd` only | 9,684 B | 9,684 B |

Reporter consumers pay about 15 percent more. Non-reporter consumers pay nothing, so tree-shaking is intact. The growth is the two new classes, the child-reporter scaffolding (`pid`, `epoch`, the snapshots and requestTargets maps, `syncPlayback`), the fixes, and some dead code (the stamps fallback). Worth a sentence in this entry so adopters see it.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -798,59 +560,30 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
 		}
 
 		// The session's sid is stamped over any per-call value here so a
-		// transform sees the session identity the report will carry.
+		// transform sees the session identity the report will carry. The store
+		// is resolved per call: a transform that rotates the session mid
+		// fan-out leaves later targets reading the archived snapshot.
 		const item: Cmcd = {
-			...session.data,
+			...(session.snapshots.get(CMCD_ROOT_PID) ?? this.playback.data),
```
littlespex, line 573: This reads no `pid` and falls back to the live root store. The plan (`architecture.md`, "Player-aware attribution") specifies `session.snapshots.get(record.pid ?? CMCD_ROOT_PID)` for an archived session and the live store only for the current one. The child-reporters RFC (#398, v5, line 449) already says this PR landed "the pid-with-root-fallback resolution rule". It did not.

Today this is wire-neutral: `rotate()` always archives the root snapshot, so the `??` only fires for the current session. The problem is the direction of the fallback. If the children PR substitutes `record.pid ?? CMCD_ROOT_PID` into this expression, a never-archived child's late response reads the root's live store, which both the plan (line 133) and the RFC (line 445) rule out. No test pins the fallback direction.

Two options: implement the pid-aware lookup now (branch on `session === this.ledger.current`, then `snapshots.get(pid) ?? snapshots.get(CMCD_ROOT_PID)`), or correct the RFC sentence. The fix commits take the second option and leave this code as is.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -501,18 +231,38 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
+	 * pinning survives a transform that rotates mid fan-out.
+	 */
+	private emitIntervalTick(session: CmcdSessionState<C>, config: CmcdEventReportConfigNormalized<C>): void {
+		const target = session.eventTargets.get(config)
+
+		if (!target) {
+			return
+		}
+
+		this.fanOutDepth++
```
littlespex, line 253: The deferred-eviction recipe (depth++, try/finally, drain, `maybeEvict()`) is hand-copied at three entry points: `start()`, here, and `emitEvent()`. The plan's Task 3 wrapped only the first and last. The tick was found by a test and fixed in 704089bd5, and its copy still differs from `emitEvent()` (no catch), which is the eviction gap on line 262. The architecture goal was to make these orderings structural. One private `runFanOut(session, fn)` that owns the depth, the drain, the eviction and the catch-then-rethrow would do that, and `requestEvict()` then collapses into `maybeEvict()`. Follow-up, not blocking.

libs/cmcd/src/CmcdOutbox.ts
```diff
@@ -0,0 +1,86 @@
...
+		this.disposed = true
+		this.queue.length = 0
+	}
+
+	process(drain: boolean): boolean {
+		if (this.disposed || !this.queue.length) {
+			return false
+		}
+
+		if (this.queue.length < this.batchSize && !drain) {
```
littlespex, line 49: Dirty tracking ended up maintained from four sites: the push in `queueTargetEvent`, this below-batch re-mark, the remainder re-mark in `processEventTargets`, and the failed-send re-queue. The identity guard in `markDirty()` and the purge in `rotate()` were added by 16a65b135 and dd345514f after the plan's design lost partial batches and resurrected batches from evicted or replaced sessions. `architecture.md` line 147 still describes the original design ("a pass that empties a session's outboxes unmarks"). Anyone building the children PR from `steps.md` as written gets the stranded-partial-batch bug back.

Two things: update `architecture.md` to what shipped (in the fix commits), and consider one channel for "still holds lines" (call `onDirty()` on exit whenever `queue.length > 0`, return only "dispatched with remainder") so the facade's re-mark disappears.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -870,26 +603,17 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-		// msd may only ride a report through the once-per-target gate, and only
-		// when the target's key filter will retain it (msd is never force-added
-		// at encode time): consuming the gate for a report that filters msd out
-		// would silently drop it for the session. A value smuggled in via
-		// per-call data or a transform is stripped, so the gate stays the
-		// single source of once-per-session semantics.
-		if (!isNaN(session.msd) && !target.msdSent && config.enabledKeys?.includes('msd')) {
-			report.msd = session.msd
+	private queueTargetEvent(session: CmcdSessionState<C>, target: CmcdEventTargetState, config: CmcdEventReportConfigNormalized<C>, report: Cmcd, type: CmcdEventType): void {
+		const attach = stampReport(report, session, target, config.enabledKeys?.includes('msd') ?? false, type)
```
littlespex, line 607: Event mode decides the `msd` gate from `enabledKeys.includes('msd')` and then calls `encodeCmcd`, which prepares and encodes in one step and discards the prepared object. Request mode checks the prepared output. The counters design (`plans/cmcd-session-counters/design.md`, line 55) needs the prepared-output rule for every target's `bsd` cursor, so the counters PR would have to rewrite this site.

`encodeCmcd` is exactly `encodePreparedCmcd(prepareCmcdData(...))`. Inlining the pair here and committing on `prepared.msd !== undefined` is byte-identical on the wire for V2 targets, and it removes the `attachMsd` mode flag from `stampReport`. For a JS-only `version: 1` event target the two checks differ only in the flag, since `msd` never reaches a V1 wire either way. Not blocking.

libs/cmcd/src/createCmcdReporterConfig.ts
```diff
@@ -0,0 +1,81 @@
...
+		// Apply target config defaults
+		eventTargets: eventTargets.reduce((acc, target) => {
+			if (target?.url && target.events?.length) {
+				acc.push({
+					version: target.version || CMCD_V2,
+					enabledKeys: target.enabledKeys?.slice() || [],
+					url: target.url,
+					events: target.events.slice(),
+					interval: target.interval ?? CMCD_DEFAULT_TIME_INTERVAL,
+					batchSize: target.batchSize || 1,
```
littlespex, line 74: Pre-existing, but this is the natural place to clamp. `batchSize || 1` lets `0.5` or `-1` through. `splice(0, 0.5)` removes nothing, `process()` POSTs an empty body and returns true, and `processEventTargets` recurses until the stack overflows. Probe on both builds: one `recordEvent` with `batchSize: 0.5` fires several thousand empty POSTs and then `RangeError: Maximum call stack size exceeded` escapes from `recordEvent`. `Math.max(1, Math.floor(batchSize))`, or an error that names the parameter and the received value.

libs/cmcd/src/CmcdOutbox.ts
```diff
@@ -0,0 +1,86 @@
...
+	constructor(url: string, batchSize: number, requester: CmcdRequester, onGone: () => void, onDirty: () => void) {
+		this.url = url
+		this.batchSize = batchSize
+		this.requester = requester
+		this.onGone = onGone
+		this.onDirty = onDirty
+	}
+
+	/**
+	 * Does not gate on `disposed`: a re-queue landing after disposal (see
```
littlespex, line 31: Three convention items across the new files.

TSDoc: seven new exports have no doc block (`createCmcdReporterConfig.ts` lines 15, 19, 25, 31, 43, `CmcdOutbox.ts` line 4, `CmcdSessionState.ts` line 10), and none of the ten new files carries the `@internal` tag every existing internal module in `libs/cmcd/src` has.

Em dashes in new prose: `stampReport.ts` lines 10, 12, 21, 23 and `acceptStateChange.ts` line 95.

New rationale comments, which the plan's rules exclude ("rationale goes in the commit message"): `CmcdReporter.ts` 185 to 187, 360 to 363, 511 to 513, 569 to 571, `CmcdOutbox.ts` 31 to 34 and 58 to 60, `CmcdSessionLedger.ts` line 4. This one on `push()` is also wrong about the code: the re-queue goes through `queue.unshift()` in `process()`, never through `push()`. `push()` needs no guard because its only caller checks `disposed` first. Two moved docs drifted too: `CmcdSessionLedger.ts` line 12 points to `startSession()` for logic that now lives in `rotate()`, and `createCmcdSessionState`'s doc promises a "dedup baseline for the default request target" that `CmcdTargetStamps` does not hold.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -1132,44 +837,27 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-		// is consumed after preparation, and only if the key filter retained
-		// msd, so a filtered-out msd is not silently dropped for the session.
-		const sendMsd = !isNaN(session.msd) && !session.requestTarget.msdSent
-
-		if (sendMsd) {
-			cmcdData.msd = session.msd
-		}
-		else {
-			delete cmcdData.msd
+		if (!stamps) {
```
littlespex, line 854: This fallback can never run: `createCmcdSessionState` seeds the default entry and nothing deletes from `requestTargets`. It leaves three copies of `{ sn: 0, msdSent: false }` (`CmcdSessionState.ts` line 66, here, and line 128) that the counters work must update in lockstep, and the dead one cannot be caught by a test. Keep one initializer and make `stamps` a `const`.

libs/cmcd/src/copyReportValues.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Copies an `SfItem`-shaped value and its `params` record.
+ *
+ * The prototype is preserved because `prepareCmcdData`, the formatter map,
+ * validation, and the structured-field encoder all branch on
+ * `instanceof SfItem`. A plain spread (and `structuredClone`) would return a
+ * prototype-less object and silently change what goes on the wire.
+ */
+function copyItemValue(value: unknown): unknown {
```
littlespex, line 35: `copyItemValue` repeats the first three branches of `copyBareValue` (`Uint8Array`, `Date`, prototype-preserving shell copy), and the outer loop in `copyReportValues` repeats the array/object dispatch a third time. The Date fix in b0a19c876 had to be written twice, which is the drift this invites, and the Buffer fix above touches two sites for the same reason. One recursive copy (primitive: return, `Uint8Array`: `new Uint8Array`, `Date`: clone, array: map, object: shell copy then recurse own properties) gives identical output, since the serializer discriminates only by `instanceof SfItem` and `instanceof SfToken`. Follow-up.

libs/cmcd/src/applyReportPolicy.ts
```diff
@@ -0,0 +1,91 @@
...
+	// transform bug. Removal and substitution are both covered, because a
+	// value the encoder drops leaves the report just as invalid as a missing
+	// one.
+	if (restoreTs && !isUsableRequiredValue(out.ts)) {
+		out.ts = ts
+	}
+
+	if (requiredKey && isUsableRequiredValue(requiredValue) && !isUsableRequiredValue((out as Record<string, unknown>)[requiredKey])) {
+		Object.assign(out, { [requiredKey]: requiredValue })
+	}
```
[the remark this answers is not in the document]
littlespex: Confirmed with a probe against both builds: an in-place `d.br.length = 0` ships `br=()`, and `d.br.pop(); delete d.br` re-attaches the mutated list. The sequence is pre-existing on main (lines 826 to 851 there), moved verbatim. Details and a zero-cost fix (capture `requiredValue` before `copyReportValues()`) are in the review comment on line 68 of this file.

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in eb6ca5059. The drain and the eviction now run inside the `finally`, so a throwing tick still evicts. Regression test: "evicts the session a throwing time-interval transform rotated away at sessionRetention 0".

libs/cmcd/src/CmcdReporter.ts
littlespex: Fixed in eb6ca5059. `bg` is stripped from the update data before the merge, so the store is never deleted from and never carries an undefined `bg` key. Test "drops the bg key from the store when update() clears it with undefined" pins the request-mode output, and c399e8ce4 discloses the change in the changelog.

libs/cmcd/src/copyReportValues.ts
littlespex: Fixed in eb6ca5059 with `new Uint8Array(value)` at both sites and a `Buffer` variant of the detachment test.

libs/cmcd/CHANGELOG.md
littlespex: Reworded in c399e8ce4: this entry is scoped to values nested in an `SfItem` and names `Buffer` input, the bare `Date` and `Uint8Array` crash has its own entry, the eviction entry covers every retention window, and the ledger in `steps.md` records both extra changes. Still open: a test for the nested-token case.

libs/cmcd/CHANGELOG.md
littlespex: Sentence added in c399e8ce4.

libs/cmcd/src/CmcdReporter.ts
littlespex: Took the second option. RFC #398 is corrected in 49f23fe52 on its branch, and `architecture.md` now says the refactor reads the root snapshot only and leaves the `pid` lookup to the children PR (c399e8ce4). Code unchanged.

libs/cmcd/src/CmcdOutbox.ts
littlespex: `architecture.md` now describes the shipped dirty tracking (c399e8ce4). The single-channel change to `process()` is not made. Still open as a follow-up.

libs/cmcd/src/applyReportPolicy.ts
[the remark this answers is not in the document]
littlespex: Fixed in 279a6a0a. The baseline is captured before copyReportValues() runs, so it points at the store's value, which the transform cannot reach. Two regression tests cover the in-place empty and the mutate-then-drop cases.

libs/cmcd/src/applyReportPolicy.ts
littlespex: Fixed in 279a6a0a per the 2026-09-09 decision: it rides this PR. The changelog and the plan's fix ledger record it. The isUsableRequiredValue doc no longer claims the encoder drops empty lists.

libs/cmcd/CHANGELOG.md
littlespex: The nested-token test landed in cd24ba46: an SfToken in an item's value and one in its params, mutated by the first transform, with the second report probing the store. That was the last open item in this thread.

libs/cmcd/src/createCmcdReporterConfig.ts
littlespex: Clamped in 6cd76b88 per the 2026-09-09 decision. The normalizer floors a number, and anything below 1 becomes 1, the same type-checked pattern sessionRetention uses. Two regression tests cover 0.5 and -1. Throwing was considered and rejected because the normalizer coerces silently everywhere else.

libs/cmcd/src/CmcdReporter.ts
littlespex: Removed in 7fa2f626. createTargetStamps() owns the literal, and resolveRequestTarget() is the one lookup path, storing what it creates. The eager seed stays, so the two repointed probes in the test file are untouched.

libs/cmcd/src/CmcdOutbox.ts
littlespex: Done in 7c3430fe. Every export in the ten new files has a TSDoc block with the internal tag. The em dashes this branch introduced are gone, and the ones inherited from main wait for issue #435. The seven rationale comments are removed, and both drifted docs are corrected.

libs/cmcd/src/CmcdReporter.ts
littlespex: Deferred to a follow-up PR per the 2026-09-09 decision. All four structural items land together after this PR merges. This thread stays open until then.

libs/cmcd/src/CmcdOutbox.ts
littlespex: Deferred to a follow-up PR per the 2026-09-09 decision. All four structural items land together after this PR merges. This thread stays open until then.

libs/cmcd/src/CmcdReporter.ts
littlespex: Deferred to a follow-up PR per the 2026-09-09 decision. All four structural items land together after this PR merges. This thread stays open until then.

libs/cmcd/src/copyReportValues.ts
littlespex: Deferred to a follow-up PR per the 2026-09-09 decision. All four structural items land together after this PR merges. This thread stays open until then.

libs/cmcd/src/CmcdReporter.ts
littlespex: Implemented in #452, stacked on this branch. This thread resolves when that PR lands.

libs/cmcd/src/CmcdOutbox.ts
littlespex: Implemented in #452, stacked on this branch. This thread resolves when that PR lands.

libs/cmcd/src/CmcdReporter.ts
littlespex: Implemented in #452, stacked on this branch. This thread resolves when that PR lands.

libs/cmcd/src/copyReportValues.ts
littlespex: Implemented in #452, stacked on this branch. This thread resolves when that PR lands.

Issue #424, opened 2026-09-02
xml: optimize parsing of dense self-closing elements


PascalThuet:
**Is your feature request related to a problem? Please describe.**

dash.js uses `@svta/cml-xml` to parse DASH manifests. Large live/DVR manifests can contain thousands, or in stress cases tens of thousands, of `SegmentTimeline` `<S/>` elements. These elements are usually self-closing and contain only a few attributes, but each one still goes through the full generic node and attribute parsing path.

Because `parseXml()` is synchronous in dash.js, this work blocks the main thread on every manifest refresh. The dash.js maintainers asked for this optimization request to be moved upstream from [Dash-Industry-Forum/dash.js#4984](https://github.com/Dash-Industry-Forum/dash.js/issues/4984#issuecomment-4117782166).

The parser improvements released in `@svta/cml-xml` 1.1.0 already help. I reran an isolated benchmark against the current 1.1.6 package to avoid relying only on the older 1.0.1 measurements.

Benchmark environment and workload:

- Node.js 20.20.2, macOS arm64
- synthetic 2,813,392-byte MPD
- 50 Periods x 4 AdaptationSets x 500 `<S/>` elements
- 100,000 `<S/>` elements in total
- 4 warm-up iterations and 20 alternating measured iterations

| Version | Median | Average | p95 |
| --- | ---: | ---: | ---: |
| `@svta/cml-xml` 1.0.1 | 31.7 ms | 33.0 ms | 47.7 ms |
| `@svta/cml-xml` 1.1.6 | 29.0 ms | 29.7 ms | 39.5 ms |

This synthetic workload intentionally stresses dense timelines and is not meant to represent every MPD. It shows that the 1.1.x improvements reduce the cost, while parsing dense self-closing elements remains a meaningful synchronous workload on desktop and is likely more visible on Smart TV/STB-class CPUs.

Typical hot-case nodes are:

```xml
<S t="123456" d="180000" />
<S d="180000" r="14" />
<S d="180000" r="-1" k="3" />
```

**Describe the solution you'd like**

Please consider optimizing `parseXml()` for documents containing many self-closing elements, while keeping the parser generic and preserving its current output shape.

Possible directions to evaluate, rather than a prescribed implementation:

- a generic fast path for self-closing elements;
- avoiding `unescapeHtml()` when a name or value contains no entity marker;
- reducing temporary strings or other per-node work;
[...]

PR #427, opened 2026-09-02
fix(xml): stop parseXml hanging on truncated declaration or close tag

littlespex:
## Description

`parseXml()` could loop forever on a document that ends in the wrong place. PR #425 fixed the attribute case. This PR fixes the two remaining cases and adds a test guard so a future hang fails CI instead of stalling it.

Fixed hangs:

- Unterminated XML declaration, for example `<?xml version="1.0" encoding="UTF-8"` with no `>`. `indexOf` returned -1, `pos` became 0, and parsing restarted from the top of the document forever. The declaration opens virtually every manifest, so a response cut off in its first few dozen bytes hung the caller's thread.
- Unterminated matching close tag, for example `<a>text</a`. `indexOf` returned -1 and the function returned with `pos` still -1. The enclosing loop then re-read the document from index 0. At the document level that looped forever. Inside a nested element it re-parsed the document as a new child on every pass until the call stack overflowed.

Both fixes consume the rest of the input when the `>` is missing, so the parser returns the partial tree it has. That matches what it already does for `<a>text` or an unterminated CDATA section. Well-formed input is unaffected because the new branches only run when `indexOf` returns -1. The previous build and this one produce identical output for the 15 XML and MPD fixtures in the repo plus 37 inline inputs, each under 4 option sets.

The new tests run `parseXml` in a worker thread. The parser is synchronous, so the node:test timeout cannot interrupt an infinite loop. The worker posts progress before each parse, and the test fails after 2 seconds of silence and names the input. The malformed-attribute tests from #425 now use the same guard. A sweep test parses every prefix of both fixtures.

The close-tag hang was not in the original report. A sweep over every prefix of a synthetic document found it, and it has its own commit.

### Notes for review

- The branch `fix/xml-throw-on-malformed-attributes` (no PR yet) rewrites the #425 attribute tests to expect throws. A dry-run merge with this branch conflicts only in `parseXml.test.ts` and `CHANGELOG.md`. Whichever lands second needs to wrap the throw expectations in the worker guard and keep both sets of changelog bullets.
- Making truncated input throw instead of returning a partial tree would be a behavior change beyond attributes. It is left for the well-formedness RFC.
- Pre-existing and out of scope: the close-tag check is a prefix match, so `</abc>` closes `<ab>`.

## Packages Changed

| @svta/cml-xml | 1.1.6 (no bump, changelog under Unreleased) | fix |

## Test Plan

- [x] `npm run build -w libs/utils -w libs/xml` then `npm test -w libs/xml`: 29 tests pass
- [x] Each new test failed before its fix (guard timeout for the two hangs, "Maximum call stack size exceeded" for the nested close tag) and passes after
- [x] `cml-xml.api.md` unchanged after the build
- [x] Every-prefix truncation sweeps of `node_types.xml`, `bbb_30fps.mpd`, a synthetic document with every construct, and a 9.6 KB cmaf-ham manifest: 0 hangs

## Requirements Checklist

- [ ] Docs/guides updated (not needed: no public API or documentation change)
- [x] Changelog updated

Refs: #424

### DIFF
--- a/libs/xml/CHANGELOG.md
+++ b/libs/xml/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 - `parseXml` no longer hangs on an attribute with no quoted value, such as `<a b>` or input truncated mid-attribute. The attribute yields an empty string and parsing resumes at the closing `>`.
 - `parseXml` no longer hangs on input that ends inside the XML declaration, such as `<?xml version="1.0"`. Parsing stops at the end of the input.
+- `parseXml` no longer hangs (or, inside nested elements, overflows the call stack) on input that ends inside a matching close tag, such as `<MPD><Period></Period`. The element keeps the children parsed so far.
 
 ## [1.1.6] - 2026-07-28
 
--- a/libs/xml/CHANGELOG.md
+++ b/libs/xml/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 ### Fixed
 
 - `parseXml` no longer hangs on an attribute with no quoted value, such as `<a b>` or input truncated mid-attribute. The attribute yields an empty string and parsing resumes at the closing `>`.
+- `parseXml` no longer hangs on input that ends inside the XML declaration, such as `<?xml version="1.0"`. Parsing stops at the end of the input.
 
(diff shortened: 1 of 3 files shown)


littlespex:

PR #428, opened 2026-09-02
fix(xml): throw on malformed attributes instead of tolerating them

littlespex:
## Description

Follow-up to #425. That PR stopped `parseXml()` from looping forever on an attribute with no quoted value, but it resolved those attributes leniently: a valueless or unquoted attribute became an empty string and parsing carried on. That is HTML behavior. In XML 1.0 an attribute is `Attribute ::= Name Eq AttValue` with `Eq ::= S? '=' S?` (§3.1, §2.3), and `AttValue` has no unquoted form. Unquoted values, valueless attributes, and a missing `=` are well-formedness violations that a conforming processor must report as fatal errors (§1.2, §2.1, §5.1).

`parseAttributes()` now walks that grammar instead of scanning forward for the next quote. Anything else throws an error that names the attribute and reports the line, column, and offending character, with `end of input` for truncated documents:

```
Malformed attribute "r": expected "=" after name
Line: 0
Column: 46
Char: end of input
```

Inputs that now throw (all hung before #425, all returned an empty-string attribute after it):

- `<a b>` and `<a b/>` (valueless attribute)
- `<a b=c/>` (unquoted value)
- `<a b="c" d>`
- `<S d="180000" r` (manifest cut off inside an attribute name)
- `<a b=` (cut off after the equals sign)

The forward scan also misattributed values across attributes. `<a b c="d"/>` assigned `"d"` to `b` and dropped `c`. The grammar walk fails at `b` instead.

One input that parsed on `main` before #425 changes behavior: `<a b "c"/>` (quoted value, no `=`) used to parse as `b="c"` and now throws. It is listed under Changed in the changelog.

The existing "Unexpected close tag" error now shares its position formatting through a small `syntaxError()` helper. Its text is unchanged. The `pos--` / `pos++` dance that #425 relied on is gone.

Well-formed input is unaffected. A well-formed attribute reaches its quote through exactly `S? = S?`, so the new walk consumes the same bytes as the old scan. I ran the parser from `main` and this branch over all 15 XML fixtures in the repo plus 14 hand-written samples (well-formed input, and malformed close tags to cover the error text), with default and full options. All 58 outputs were identical.

Tests: the malformed-attribute tests from #425 now assert the thrown message and the line and column reporting (10 tests), and the 3 well-formed guards are unchanged. Nine of the ten failed before the parser change. Root `npm test` (lint, build, typecheck, all package tests) passes, including the 47 drm tests that consume the parser.

For reviewers: `origin/issue/xml-parse-hang` (unmerged) fixes two more truncation hangs, an unterminated `<?xml` and an unterminated close tag, by returning the partial tree, and it rewrites the same malformed-attribute tests to run in a worker. The two branches conflict in `libs/xml/test/parseXml.test.ts`, and they disagree on policy: this PR throws on non-well-formed input, that one tolerates it. Whichever lands second needs its attribute tests re-aligned, and if the direction is spec compliance, those two fixes should throw through `syntaxError()` as well.

## Requirements Checklist

- [x] Docs/guides updated (`@throws` added to the `parseXml` TSDoc; no public API change, `cml-xml.api.md` is unchanged)
- [x] Changelog updated

### DIFF
--- a/libs/xml/CHANGELOG.md
+++ b/libs/xml/CHANGELOG.md
@@ -10,7 +10,11 @@ and this project adheres to
 
 ### Fixed
 
-- `parseXml` no longer hangs on an attribute with no quoted value, such as `<a b>` or input truncated mid-attribute. The attribute yields an empty string and parsing resumes at the closing `>`.
+- `parseXml` no longer hangs on an attribute with no quoted value, such as `<a b>`, `<a b=c/>`, or input truncated mid-attribute. It now throws an error that names the attribute and reports the line and column, matching the XML 1.0 `Attribute ::= Name Eq AttValue` production.
+
+### Changed
+
+- `parseXml` throws on a quoted attribute value with no `=` (`<a b "c"/>`), which previously parsed leniently as `b="c"`.
 
 ## [1.1.6] - 2026-07-28
 
--- a/libs/xml/src/parseXml.ts
+++ b/libs/xml/src/parseXml.ts
@@ -13,6 +13,11 @@ const SINGLE_QUOTE_CC = 39            // "'"
 const DOUBLE_QUOTE_CC = 34            // '"'
 const OPEN_CORNER_BRACKET_CC = 91     // '['
 const CLOSE_CORNER_BRACKET_CC = 93    // ']'
+const EQUALS_CC = 61                  // '='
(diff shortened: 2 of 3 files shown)


littlespex:

PR #431, opened 2026-09-02
[RFC] Consumer-defined XML builders

littlespex:
## Description

This RFC proposes `parseXmlWith(input, builder, options)` for `@svta/cml-xml`. It parses an XML string in one pass over the scanner from #432, and it gives each element to a *builder* that the consumer supplies. `createElement` returns the consumer's own object. That object is the `parent` for the element's children, and it is the `child` that `appendChild` receives when the element ends. Returning `undefined` skips the element. No `XmlNode` tree is built.

`parseXml` keeps its signature and output. It shares the scanner with `parseXmlWith`, and three fixes to how the scanner reads XML apply to both.

**Revised after review (2026-09-03).** The first draft proposed an incremental parser with `write` and `end`. Every blocking finding was about that part, #432 had already shipped the scanner and an internal builder, and dash.js parses complete strings. Incremental input therefore moves to its own RFC. Its design notes stay in the design record. The file keeps its working name.

Files:

- `rfc/xml-incremental-parser.md`: the proposal. The builder contract, strict end-of-input handling, the grammar fixes, measurements against `main` after #432, drawbacks, alternatives, and open questions.
- `plans/xml-incremental-parser/prototype/`: an executable TypeScript prototype. Node 24 runs it against `main`, and the repository's lint and typecheck cover it. It contains the scanner with the proposed changes, `parseXml` on that scanner, `parseXmlWith`, a faithful port of dash.js `processNode` next to a one-pass builder that produces identical objects, the equivalence script, and the benchmark.
- The other files in `plans/xml-incremental-parser/`: the design record. Findings with the review decisions and the deferred incremental design, equivalence results, benchmark results, implementation steps, and the flattening builder example that the RFC links to.

Measured on the prototype (one process per variant, natural GC, medians, Node 24.16 on an M1 Pro on battery):

| Input | `parseXml` on `main` | `parseXml`, phase-one scanner | dash.js today on `main` | One pass, faithful | One pass, lean |
|---|---:|---:|---:|---:|---:|
| 100k `<S>`, pretty, 3.7 MB | 22.7 ms | 23.1 ms | 41.4 ms | 37.7 ms | 32.9 ms |
| 100k `<S>`, minified, 2.6 MB | 16.4 ms | 16.7 ms | 35.8 ms | 33.5 ms | 30.6 ms |
| livesim2 real, 170 KB | 0.821 ms | 0.804 ms | 1.38 ms | 1.43 ms | 1.30 ms |

Notes for reviewers:

- The contract changes cost `parseXml` nothing measurable. For dash.js, a lean builder gains 13 to 21 percent on the two large synthetic inputs and 6 percent on the real manifest, on top of #432. A faithful one-pass replacement of `processNode` gains almost nothing. The RFC says so under Measured results and Drawbacks. Its case is the consumer-shaped output, skipping, and strictness, not the 2x from #424. #432 already delivered most of that 2x.
- The grammar fixes change `parseXml` output on eight parity-corpus inputs. `equivalence.md` lists each one with its reason.
- A `parseXml`-only bundle grows by about 450 bytes gzipped. See Bundle size in the RFC.
- One result is unexplained. With a full GC before every call, the faithful builder runs 4 to 5 times slower on the 150 KB inputs. It does not when it alternates with `parseXml`. The RFC records this as an open question for the implementation to settle with `--trace-deopt`.
- `this: void` on the callbacks needs `allowAsThisParameter` in the ESLint configuration. The prototype has a file-level disable until the implementation adds it.

This PR changes no package code. The public API changes only if an implementation PR follows acceptance.

## Packages Changed

| none (docs: `rfc/`, `plans/`) | n/a | docs |

## Test Plan

- [x] `node plans/xml-incremental-parser/prototype/equivalence.ts`: 778 checks, 0 failures, 34 expected differences, against `main` at 5e4a82276
- [x] `npm run typecheck` and `npm run lint` pass with the prototype included
- [x] The six TypeScript examples in the RFC and the flattening example in `plans/xml-incremental-parser/flattening-builder.md` typecheck under `strict` against the prototype's `index.ts`
- [x] `node plans/xml-incremental-parser/prototype/bench.ts` with and without `--gc`; results in `benchmark.md`
- [x] Markdown and prototype only; the build, the package tests, and the docs build are not affected

## Requirements Checklist

- [ ] Unit Tests updated or fixed (not applicable: RFC and design record only)
- [x] Docs/guides updated (the RFC is the document)
- [ ] Changelog updated (not applicable: no package change)

Refs: #424, #426, #430, #432, [dash.js#4984](https://github.com/Dash-Industry-Forum/dash.js/issues/4984)


rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,417 @@
...
+```ts
+import { XmlParser, type XmlBuilder } from '@svta/cml-xml'
+
+type Node = Record<string, unknown> & { tagName: string; children: Node[]; text?: string }
+
+const builder: XmlBuilder<Node> = {
+	createDocument: () => ({ tagName: '#document', children: [] }),
+	createElement: (parent, name, attributes, localName) => ({ tagName: localName, children: [], ...attributes }),
+	appendChild: (parent, child) => { parent.children.push(child) },
+	appendText: (parent, text) => { parent.text = text },
```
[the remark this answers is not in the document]
littlespex: Fixed in d8d7a6a. The Summary example now accumulates text: `parent.text = (parent.text ?? '') + text`. The dash.js example keeps only the last run, on purpose. `processNode` assigns `node.__text = child.nodeValue` for each text child today. The RFC now says so next to that example.

plans/xml-incremental-parser/benchmark.md
```diff
@@ -0,0 +1,564 @@
...
+```js
+// Benchmark for the xml-incremental-parser prototype.
+//   node bench.mjs                 driver: one process per (variant, input) so call sites stay monomorphic
+//   node bench.mjs --interleaved   all variants in one process (shows the megamorphic penalty)
+//   add --gc to force a full GC before every timed call (driver passes --expose-gc to workers)
+import { readFileSync } from 'node:fs'
+import { performance } from 'node:perf_hooks'
+import { spawnSync } from 'node:child_process'
+import { fileURLToPath } from 'node:url'
+import { parseXml as shippedParseXml } from '/Users/cocch1216/dev/common-media-library/.claude/worktrees/streaming-api-rfc-xmlnode-7d96c7/libs/xml/dist/index.js'
```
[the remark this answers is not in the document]
littlespex: Fixed in d8d7a6a. The harness imports `../../libs/xml/dist/index.js` and reads fixtures from `../../libs/xml/test/fixtures/`, both relative to `plans/xml-incremental-parser/`. The Harness section now explains where to put the files and the downloaded livesim2 manifest. `prototype.md` includes the measurement-only variant as a diff, so the scripts run as written.

plans/xml-incremental-parser/equivalence.md
```diff
@@ -0,0 +1,301 @@
...
+`<![CDA`, truncated comment, truncated CDATA, truncated doctype, attribute names starting with a digit
+or underscore (skipped by the grammar), and non-ASCII text.
+
+## Script
+
+```js
+// Equivalence and chunk-boundary checks for the xml-incremental-parser prototype.
+import assert from 'node:assert'
+import { readFileSync } from 'node:fs'
+import { parseXml as shippedParseXml } from '/Users/cocch1216/dev/common-media-library/.claude/worktrees/streaming-api-rfc-xmlnode-7d96c7/libs/xml/dist/index.js'
```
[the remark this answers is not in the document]
littlespex: Fixed in d8d7a6a. The script uses the same repository-relative paths as the harness. A note above the script says where to run it from.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,417 @@
...
+- `appendComment(parent, text)` receives the whole comment including `<!--` and `-->`, matching the `#comment` node value `parseXml` produces with `keepComments`. Absent handler, comments are skipped.
+- `appendDoctype(parent, text)` receives the text from `!` to before the closing `>`, honoring a bracketed internal subset, matching `#doctype` nodes.
+- The XML declaration and processing instructions are skipped, as today.
+- Builder functions must not call `write` or `end` on the parser that is invoking them.
+
+### Incremental tokenization
+
+`write(chunk)` prepends the carried tail from the previous call, if any, and scans the result in streaming mode: a construct is consumed only when it is complete. Complete means the closing `>` of a start tag as found by the attribute scanner (a `>` inside a quoted value does not count), the `>` of a close tag, `?>` for a declaration or processing instruction, `-->` for a comment, `]]>` for CDATA, the `>` outside `[...]` for a doctype, and the next `<` for a text run. The first incomplete construct stops the scan; everything from its start is carried into the next `write`.
```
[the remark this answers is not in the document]
littlespex: Correct, the sentence was wrong. Fixed in d8d7a6a. Declarations and processing instructions end at the next `>`, which is how `parseXml` scans them today (`indexOf('>')`). The RFC now also notes that a `>` inside processing-instruction data ends the construct, and that this behavior is kept.

littlespex (commented):
Notes from trying the builder contract against `parseDash` in the avia-js CMAF adapter. That adapter converts a DASH MPD directly into a HAM-shaped model and keeps no tree. The numbers below come from that adapter on the livesim2 fixture that merged with #432.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+	write(chunk: string): this;
+	end(): T;
+}
+```
+
+`T` is inferred from the builder. `write` returns the parser, so the one-shot form reads `new XmlParser(b).write(s).end()`.
+
+### Callback semantics
+
+- The parser calls `createDocument` once, in the constructor. Its result is the parent of every top-level element, text run, doctype, and comment, and `end()` returns it.
```
littlespex: `createDocument` takes no arguments and runs in the constructor. A builder therefore has no way to receive per-parse input. Our DASH builder needs the request's base URL, a logger, and two running counters. Line 345 says builders should be module-level constants, so a closure per parse is not an option. Mutable module-level state breaks as soon as two time-sliced parses interleave, and the example at line 159 makes that likely (two live handlers refreshing at the same time).

Suggestion: add an option that supplies the root value, `new XmlParser(builder, { root })`. `end()` returns that value, and `createDocument` becomes optional. The merged scanner (#432) already accepts a pre-seeded stack, so this change is in the wrapper only.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+
+```ts
+export type XmlParserOptions = {
+	keepWhitespace?: boolean;
+};
+
+export type XmlBuilder<T> = {
+	createDocument(): T;
+	createElement(parent: T, name: string, attributes: Record<string, string>, localName: string, prefix: string | null): T;
+	appendChild?(parent: T, child: T): void;
```
littlespex: Consider passing the element name, or `localName`, as a last argument to `appendChild` and `appendText`. The scanner has the name on the `names` stack at both call sites, so the cost is one argument. Without it, every value a builder returns must carry its own tag name, so that `appendChild` can tell what it received. The dash.js example stores `tagName` on every node for exactly this reason. With the name passed in, a builder can return a plain `{ t, d, r }` for `<S>` and handle it by name. As a last argument, it keeps two-argument builders valid.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+| **Author** | Casey Occhialini |
+| **Date** | 2026-09-02 |
+| **Package** | `@svta/cml-xml` |
+| **Breaking change** | No |
+| **Tracking** | [#424](https://github.com/streaming-video-technology-alliance/common-media-library/issues/424), [dash.js#4984](https://github.com/Dash-Industry-Forum/dash.js/issues/4984) |
+| **Design record** | `plans/xml-incremental-parser/` |
+
+## Summary
+
+Add an incremental parser, `XmlParser`. It tokenizes XML and hands each element, text run, CDATA section, comment, and doctype to a *builder* that the consumer supplies. It assembles no `XmlNode` tree. The builder decides what an element becomes: whatever `createElement` returns is the parent that the element's children receive, and the child that `appendChild` receives when the element completes. A manifest parser can therefore produce its final objects in the single tokenization pass, with no intermediate tree and no second walk.
```
littlespex: This is true for a tree-shaped target like the dash.js object, where inheritance is resolved afterwards. It is not true for a builder that flattens, which is what a HAM converter does. In most manifests the `SegmentTemplate` is on the AdaptationSet, and its `<S>` children arrive before the Representations that share it. A segment needs the Representation id and bandwidth for its URL, so `<S>` cannot become a segment in `createElement`. The builder keeps a small record per `<S>` and expands the records in `appendChild`, when the Representation or the Period ends.

That is still one pass, but the guide should describe this deferral. It also makes the builder depend on document order. The XSD sequence puts `SegmentTemplate` before `Representation` and `BaseURL` before `Period`. Today's tree consumers do not depend on that. One data point on the gain: in our adapter, the tree is about 9 percent of manifest parse time, and URL templating is about 60 percent.

Suggestion: shorten this sentence to "with no intermediate tree", and add a short section on inheritance-resolving builders to the guide.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+- The parser calls `appendChild(parent, child)` when the element completes: at its close tag, directly after `createElement` for a self-closing tag, or during `end()` for an element the input never closed. Siblings complete in document order, so post-order appends preserve child order.
+- The parser calls `appendText(parent, text)` once per text run, when the run completes. With `keepWhitespace` unset (the default), the text is entity-decoded and trimmed, and runs that are empty after trimming are not delivered. With `keepWhitespace: true`, the text is entity-decoded and delivered as is when non-empty. Text at document level goes to the document value.
+- `appendCdata(parent, text)` receives the raw content between `<![CDATA[` and `]]>`, not decoded.
```
littlespex: These three callbacks deliver inconsistent shapes. The comment includes `<!--` and `-->`. The doctype includes the `!` but not the `<` and `>`. CDATA excludes its wrapper. They match `XmlNode` because `parseXml` needs parity, but the builder layer exists to shape output. The tree builder can add back whatever `XmlNode` expects, and the new API could deliver the inner text in all three cases.

The same applies to attribute names at line 243. A name that does not start with an ASCII letter is skipped one character at a time, so `<a _x="1">` yields `x="1"`. This is inherited and acceptable for `parseXml`, but a new public contract can define it correctly. Both changes are cheap while `XmlBuilder` is still internal. After release, they are fixed.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+The same design fixes two more problems of a whole-string synchronous parser. A 25 ms parse is one long task that delays input handling at every refresh, and on TV-class CPUs the same parse takes several times longer. And a manifest that is still downloading cannot be parsed until the last byte arrives, although tokenization could run behind the network.
+
+## Guide-level explanation
+
+### The builder contract
+
+A builder is a plain object of functions. The parser calls `createDocument` once. Then it calls `createElement` for every start tag, with the value that the enclosing element's `createElement` returned as `parent`. Whatever `createElement` returns becomes the parent for that element's children, and `appendChild` receives it when the element completes.
+
+```ts
+export type XmlBuilder<T> = {
```
littlespex: The merged `XmlBuilder.ts` (#432) uses property syntax: `createElement: (...) => T`. This block uses method syntax. Under `strictFunctionTypes`, only property syntax checks parameter types strictly (contravariantly). A builder for a union `T` that declares a narrower `parent` type is rejected with property syntax and accepted with method syntax. The implementation has the safer form. Align the RFC with it.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Unresolved questions
+
+1. Names: `XmlBuilder` with `createElement` and `appendChild` (proposed), or `XmlHandlers` with `startElement` and `endElement`?
+2. Class or factory: `new XmlParser(builder)` (proposed, matches `WebVttParser`), or `createXmlParser(builder)` (matches `createFetchTransport`)? Performance is equal as long as the scanner is module-level.
+3. A Web Streams wrapper: ship an `XmlTransformer` next to the parser, mirroring `WebVttTransformer`, or leave `pipeTo(new WritableStream(...))` as the documented pattern? Proposed: leave it out until someone asks.
+4. The whitespace option: does `keepWhitespace` belong on `XmlParserOptions`, or should `appendText` always receive the raw run and let the builder trim? Proposed: keep the option. Trimmed non-empty text is what almost every consumer wants, and `parseXml` needs the same rule.
+5. Processing instructions: add `appendProcessingInstruction`? `parseXml` skips them today. Proposed: not in this RFC.
+6. Error type: a plain `Error` with an `Offset` line (proposed, matches today), or an `XmlParseError` subclass with an `offset` field?
```
littlespex: From the adapter side: use the subclass with an `offset` field. Adapters wrap parser errors in their own error types, and they should read a field instead of parsing a message. On the other questions: keep `keepWhitespace` as an option (question 4), and skip the streams wrapper and processing instructions until someone asks for them (questions 3 and 5).

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Future possibilities
+
+- An `XmlTransformer` for `TransformStream` pipelines.
+- A generic "objectify" builder in `@svta/cml-dash` (named properties, configurable array elements, converted attributes), so players get an MPD object without writing a builder.
```
littlespex: An objectify builder reproduces only the dash.js shape. The HAM package in this repository is the other natural consumer. It exercises the parts that the dash.js example never touches: inheritance, expansion to segments, and deferred completion. A HAM builder as the second reference consumer would test the API harder than a second tree shape would.

plans/xml-incremental-parser/steps.md
```diff
@@ -0,0 +1,42 @@
...
+(truncation fixes, malformed attributes and mismatched close tags throw) is merged, so the parity corpus
+is captured against the parser as it will ship. Inputs that #430 turns into errors (`<a b>`, `<a b=c/>`,
+`<a></ab>`, `<a/></b><c/>`, truncation inside an attribute) change from tree expectations to `throws`
+expectations in the corpus.
+
+1. Capture parity fixtures before touching the parser. Add `libs/xml/test/parseXml.equivalence.test.ts`
+   that parses every corpus input (`equivalence.md`) with each option set and compares to expected output
+   generated once from the current implementation. Serialize `parentElement` as a node path since JSON
+   cannot hold the cycle, or assert it separately.
+2. Add `libs/xml/src/XmlBuilder.ts` (type), `libs/xml/src/XmlParserOptions.ts` (type), and
```
littlespex: Steps 1 to 3 were written before #432, which has merged. The parity corpus exists, `scan` exists, and `parseXml` calls `scan` directly instead of going through `XmlParser`. What remains is adding streaming mode to `scan` (the `final` flag and its early returns) and the `XmlParser` wrapper. Please rewrite the steps against that base. Also add a step to re-run the #432 benchmark after the streaming guards land, because they are in the same main loop.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+
+The same design fixes two more problems of a whole-string synchronous parser. A 25 ms parse is one long task that delays input handling at every refresh, and on TV-class CPUs the same parse takes several times longer. And a manifest that is still downloading cannot be parsed until the last byte arrives, although tokenization could run behind the network.
+
+## Guide-level explanation
+
+### The builder contract
+
+A builder is a plain object of functions. The parser calls `createDocument` once. Then it calls `createElement` for every start tag, with the value that the enclosing element's `createElement` returned as `parent`. Whatever `createElement` returns becomes the parent for that element's children, and `appendChild` receives it when the element completes.
```
littlespex: What happens when `createElement` returns `undefined`? From reading the scanner: the children get `undefined` as parent, `appendChild(parent, undefined)` is called, and nothing breaks. Either document that as the way to ignore an element, or make it a real skip, where the scanner stops calling the builder until the matching close tag. The skip costs one branch per element, so measure before you choose.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+};
+```
+
+`name` is the tag name as written (`tt:span`), `localName` the part after the prefix (`span`), and `prefix` the part before it (`tt`) or `null`. Without a prefix, `localName` is the same string as `name` and nothing extra is allocated. A builder that does not care about namespaces declares three parameters and ignores the rest.
+
+`appendChild` runs when the child is complete, after all of the child's own children are appended, so it is post-order. A consumer that wants the child visible in the parent from the start attaches it inside `createElement`, since the parent is right there. Text arrives once per run, after the run ends at the next `<` or at `end()`, so a run split across chunks is still one call. Defining a handler opts in: without `appendComment`, the parser skips comments and never slices them out of the input.
+
+### dash.js in one pass
+
+This replaces `parseXml` plus `processNode`. No `XmlNode` is allocated. The objects created here are the ones the rest of the player consumes. In this example, `convert` replaces the dash.js matcher chain and `ARRAY_NODES` its `arrayNodes` list.
```
littlespex: **[P1] Test the guide's builder before calling it a replacement.** This example is not equivalent to the current `DashParser.processNode`:

- `ARRAY_NODES` contains only six of dash.js's roughly thirty array-valued elements.
- `convert` omits the duration, date-time, and language matchers.
- The `<S>` branch drops `n` and every unlisted or extension attribute.
- Omitting CDATA and non-element children changes `__children`.

The equivalence script imports a different `builders.mjs` and compares a reduced canonical form, so it does not prove parity for this copy-paste example. Please do one of two things. Either make this exact example faithful and test it, with repeated `Location`, `Event`, and `Role` elements, with duration, date, and language values, with `S@n`, with extension attributes, and with CDATA and mixed content. Or label it illustrative and remove the replacement and parity claims.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+- the next `<` for a text run.
+
+The first incomplete construct stops the scan. Everything from its start is carried into the next `write`.
+
+`end()` scans the carry in final mode, where the end of input terminates every construct exactly as `parseXml` treats a truncated document today. It then completes still-open elements from the innermost outwards through `appendChild`, and returns the document value.
+
+One scanner runs both modes. So `new XmlParser(b).write(s).end()` produces the same callbacks as feeding `s` in any split, and `parseXml(s)` on the new scanner produces the same tree as today. Both were verified on the prototype (see Testing).
+
+The carry is bounded by the size of one construct: one tag, one text run, one comment. In a manifest that is a few dozen bytes. On the 3.7 MB stress manifest split in 64 KB chunks, the largest carry was 30 characters. A construct that spans many chunks is rescanned from its start on each `write`. A document with a multi-megabyte text node therefore still parses correctly but costs more than a whole-string parse. That shape does not occur in manifests, so the RFC notes it and does not optimize for it.
```
littlespex: **[P1] This carry strategy is quadratic for a long construct.** Every `write()` flattens and rescans the whole incomplete text run, quoted attribute, comment, or CDATA section. An 8 MiB construct that arrives in 64 KiB chunks causes more than 500 MiB of scanning and string copying in total. Smaller network chunks make it worse. Streamed input is controlled by the remote side, so this is also an algorithmic denial-of-service path. Keep lexical progress across writes, or buffer the fragments without rescanning and join them once when the callback fires. Please add chunk tests with multi-megabyte text, attributes, comments, and CDATA, and consider a limit on the size of a buffered construct.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+- The parser calls `createElement(parent, name, attributes, localName, prefix)` when it has read a start tag completely, before any of the element's content. `name` is the tag name as written. `localName` and `prefix` split it at the first colon. Without a colon, `localName === name` and `prefix === null`. `attributes` is a fresh object per element with values already entity-decoded. The consumer owns it and may keep or mutate it. Malformed attributes are errors, exactly as in `parseXml` since #430.
+- The parser calls `appendChild(parent, child)` when the element completes: at its close tag, directly after `createElement` for a self-closing tag, or during `end()` for an element the input never closed. Siblings complete in document order, so post-order appends preserve child order.
+- The parser calls `appendText(parent, text)` once per text run, when the run completes. With `keepWhitespace` unset (the default), the text is entity-decoded and trimmed, and runs that are empty after trimming are not delivered. With `keepWhitespace: true`, the text is entity-decoded and delivered as is when non-empty. Text at document level goes to the document value.
```
littlespex: **[P1] The new builder API should not discard character data by default.** Trimming each run turns `<p>Hello <b>x</b> world</p>` into `"Hello"`, `"x"`, and `"world"`. The significant spaces are lost. A builder that implements `appendText` but omits the optional `appendCdata` also loses CDATA content, silently. Prefer delivering untrimmed, entity-decoded text to custom builders, and let the existing `parseXml` tree builder do its own trimming. If `appendCdata` is absent, fall back to `appendText`. That preserves the content, and a builder that supplies `appendCdata` can still tell the two forms apart.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+- the `>` of a close tag,
+- the next `>` for a declaration or processing instruction (`parseXml` does not look for `?>`, so a `>` inside processing-instruction data ends the construct today, and that stays),
+- `-->` for a comment,
+- `]]>` for CDATA,
+- the `>` outside `[...]` for a doctype,
+- the next `<` for a text run.
+
+The first incomplete construct stops the scan. Everything from its start is carried into the next `write`.
+
+`end()` scans the carry in final mode, where the end of input terminates every construct exactly as `parseXml` treats a truncated document today. It then completes still-open elements from the innermost outwards through `appendChild`, and returns the document value.
```
littlespex: **[P1] A new streaming parser should not treat a truncated response as valid XML by default.** If an HTTP body ends cleanly in the middle of a document, this rule closes the open elements and returns a partial manifest that looks valid. Compatibility with the existing `parseXml` can stay as an internal option for the final mode. But `XmlParser.end()` should reject unfinished constructs and unclosed elements by default. Otherwise, callers cannot tell a complete document from a truncated transfer.

plans/xml-incremental-parser/benchmark.md
```diff
@@ -0,0 +1,570 @@
...
+| dash.js one pass (builder) | 0.034 ms | 0.039 ms | -18.2% |
+| dash.js one pass, tuned builder (S branch, Set) | 0.034 ms | 0.038 ms | -18.2% |
+| dash.js one pass, attribute callbacks variant | 0.034 ms | 0.040 ms | -16.4% |
+| dash.js one pass, 64 KB chunks | 0.034 ms | 0.037 ms | -18.2% |
+| dash.js one pass, 16 KB chunks | 0.033 ms | 0.037 ms | -19.0% |
+| dash.js one pass, tuned builder, 64 KB chunks | 0.034 ms | 0.038 ms | -18.3% |
+
+## Harness
+
+The code blocks below and the core in `prototype.md` are saved as `generate.mjs`, `bench.mjs`, `xmlParser.mjs`
```
littlespex: **[P1] The runnable evidence is still incomplete.** This PR contains only Markdown under this folder. The four `.mjs` files named here and `livesim2_tsbd21600.mpd` are missing. The embedded code blocks also never define several imports from `builders.mjs` (`dashCurrent`, `canonicalDash`, `parseXmlChunked`, `createRecordingBuilder`, and others), and the promised livesim2 URL and checksum are not given. #432 has now merged with a real fixture, a parity corpus, and a benchmark harness. Please commit the incremental variants as executable files against that base, pin the fixture hash and the baseline commit, and rerun the results. Scratch files that nobody can audit are not enough.

plans/xml-incremental-parser/benchmark.md
```diff
@@ -0,0 +1,570 @@
...
+function fmt(ms) {
+	return ms >= 10 ? ms.toFixed(1) : ms >= 1 ? ms.toFixed(2) : ms.toFixed(3)
+}
+
+let sink = null
+
+function measure(variantIndexes, input) {
+	const text = input.make()
+	// Fresh flat strings per chunk, as TextDecoderStream produces them (slices would be SlicedStrings).
+	const flat = (parts) => parts.map(part => Buffer.from(part, 'utf8').toString('utf8'))
```
littlespex: **[P1] The whole and chunked variants do not receive comparable strings.** The synthetic whole input is built with repeated `out +=`. By this RFC's own V8 argument, that produces a rope string, which the scanner reads more slowly. The chunk path then passes every slice through a Buffer and back, which produces a flat string. That difference in string representation can explain the reported chunked win, so the result does not prove that chunking costs nothing. Flatten one common whole input first. Then benchmark separately: the flat whole input, decoder-like flat chunks, and the `slice()` chunks that the time-slicing guide uses. Include the slicing cost and the maximum duration of one write.

plans/xml-incremental-parser/benchmark.md
```diff
@@ -0,0 +1,570 @@
...
+| dash.js one pass (builder) | 0.087 ms | 0.145 ms | -48.7% |
+| dash.js one pass, tuned builder (S branch, Set) | 0.091 ms | 0.161 ms | -46.2% |
+| dash.js one pass, attribute callbacks variant | 0.075 ms | 0.129 ms | -55.8% |
+| dash.js one pass, 64 KB chunks | 0.091 ms | 0.168 ms | -46.3% |
+| dash.js one pass, 16 KB chunks | 0.086 ms | 0.137 ms | -49.5% |
+| dash.js one pass, tuned builder, 64 KB chunks | 0.084 ms | 0.141 ms | -50.5% |
+
+### All variants interleaved in one process (contamination demonstration, natural GC)
+
+One `scan` function driven by every builder and by both whole and chunked input in the same process. With the earlier prototype (state on the instance, no slots object) this inverted the result: every variant ran 30 to 65 percent slower than the shipped parser. With the final structure the relative ordering matches the per-process tables; absolute times are inflated by roughly 20 percent across the board by the garbage of fourteen variants sharing one heap, and the specialized builder loses its edge because its call sites are no longer monomorphic. Kept as the record of that effect.
```
littlespex: **[P2] The tables do not support the claim that the relative order is preserved.** On the pretty stress input, the isolated plain, tuned, and attribute-callback builders take 34.8, 29.0, and 29.6 ms. Interleaved, they take 32.8, 34.2, and 29.4 ms. The tuned builder changes from faster than plain to slower. The closed stress input reverses the order too. A real application may run `parseXml` plus one or more custom builders through the shared scanner. So the one-process-per-variant results are best cases, not the general contract. Remove the ordering claim. Add realistic two-builder runs, randomized repeated processes, and browser-engine coverage.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+
+Two things improve for free. Nesting depth is no longer limited by the call stack, because the element stack is explicit. The termination fixes for truncated input (#425 and #430) apply to both entry points, because there is one scanner.
+
+### Measured results
+
+The numbers below come from the prototype. `plans/xml-incremental-parser/benchmark.md` has the harness, the full tables, and the forced-GC run. Each variant ran in its own process: 4 warm-up calls, then 60 measured calls on the large inputs and 300 or 3,000 on the small ones. The tables show medians of natural-GC runs on Node 24.16, on an M1 Pro on battery in Low Power Mode. Treat the absolute numbers as indicative. The relative numbers are the result.
+
+Three labels recur below. "dash.js today" is the shipped `parseXml` plus a faithful port of `processNode`. "One pass" is the builder from the guide without the `<S>` branch, with the same per-node logic as `processNode`. "One pass, specialized" adds the `<S>` branch and a `Set` for the array-node lookup.
+
+The tree path: `parseXml` on the new scanner against the shipped parser, whole string and in 64 KB chunks.
```
littlespex: **[P1] Rebase the value and size analysis on the merged #432.** The flat scanner, the faster tree path, the recursion-depth removal, the parity fixtures, and much of the bundle increase now exist independently of this public API. The full benchmark already has a `dash.js on new parseXml + processNode` row. But the RFC summary compares only against the pipeline before #432, so it credits the scanner's gains to `XmlParser`. Please make the post-#432 row the decision baseline, remeasure the delta of the incremental class and builder, and keep #432's direct `parseXml -> scan` path, so that `parseXml`-only bundles can tree-shake the lifecycle wrapper.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+The parser takes input through `write(chunk)` and finishes with `end()`. The same object parses a whole string, a `fetch` body as it arrives, or a large string in slices between event-loop turns. `parseXml` keeps its signature and output, becomes one builder on top of the new scanner, and gets faster.
+
+```ts
+import { XmlParser, type XmlBuilder } from '@svta/cml-xml'
+
+type Node = Record<string, unknown> & { tagName: string; children: Node[]; text?: string }
+
+const builder: XmlBuilder<Node> = {
+	createDocument: () => ({ tagName: '#document', children: [] }),
+	createElement: (parent, name, attributes, localName) => ({ tagName: localName, children: [], ...attributes }),
```
littlespex: **[P2] XML attributes can overwrite this builder's own fields.** `...attributes` comes after `tagName` and `children`. So `<root children="x"><child/></root>` replaces the array with a string, and the next `appendChild` crashes. `tagName="shadow"` silently changes the element's identity. Keep attributes in a nested object. Or at least apply the spread before the reserved structural fields, and document what happens on a collision.

plans/xml-incremental-parser/prototype.md
```diff
@@ -0,0 +1,575 @@
...
+			catch (error) {
+				this.state = FAILED
+				throw error
+			}
+		}
+		this.carry = ''
+		const appendChild = this.slots.appendChild
+		const stack = this.stack
+		const names = this.names
+		while (stack.length > 1) {
```
littlespex: **[P2] A callback failure at end of input breaks the stated failure lifecycle.** The final `appendChild` flush is outside the `try/catch`. If a builder throws there, the stack is already popped, but `state` stays `OPEN`. A later `write()` or a second `end()` can then continue on corrupted partial state. Put the final scan and the whole flush inside the same failure transition. Set `FAILED` before rethrowing. Add a test where `appendChild` throws while an unclosed element is completed.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+- The parser skips the XML declaration and processing instructions, as today.
+- Builder functions must not call `write` or `end` on the parser that is invoking them.
+
+### Incremental tokenization
+
+`write(chunk)` prepends the tail carried from the previous call, if any, and scans the result in streaming mode: a construct is consumed only when it is complete. Complete means:
+
+- the closing `>` of a start tag as found by the attribute scanner (a `>` inside a quoted value does not count),
+- the `>` of a close tag,
+- the next `>` for a declaration or processing instruction (`parseXml` does not look for `?>`, so a `>` inside processing-instruction data ends the construct today, and that stays),
```
littlespex: **[P2] The existing `parseXml` grammar should not automatically become the new public contract.** XML processing instructions end at `?>`, not at the first `>`. The doctype scan also needs to ignore brackets and `>` inside quoted literals. For example, `<?pi a > b?><root/>` is currently split into a processing instruction plus document text. Keep the old `parseXml` output through a compatibility mode if necessary. But fix or parameterize the shared scanner for `XmlParser`, instead of knowingly making the old mis-tokenization its permanent behavior.

rfc/xml-incremental-parser.md
```diff
@@ -0,0 +1,430 @@
...
+	createDocument(): T;
+	createElement(parent: T, name: string, attributes: Record<string, string>, localName: string, prefix: string | null): T;
+	appendChild?(parent: T, child: T): void;
+	appendText?(parent: T, text: string): void;
+	appendCdata?(parent: T, text: string): void;
+	appendComment?(parent: T, text: string): void;
+	appendDoctype?(parent: T, text: string): void;
+};
+
+export class XmlParser<T> {
```
littlespex: **[P2] Separate the document and element result types before the API is fixed.** A document accumulator often has a different shape from the values returned for elements. With one `T`, consumers must use a union everywhere, and `end()` returns that union even though it always returns the document. Consider `XmlBuilder<TElement, TDocument = TElement>` and `XmlParser<TElement, TDocument = TElement>`, with `createDocument(): TDocument`, `createElement(parent: TDocument | TElement): TElement`, element children typed as `TElement`, and `end(): TDocument`. This also works well with the root-injection proposal in the other thread.

rfc/xml-incremental-parser.md
littlespex: Agreed. One addition: give each function-valued property an explicit `this: void`. The proposed implementation copies the callbacks into a fixed-shape slots object and calls them as plain functions. Without `this: void`, a contextually typed object method can use `this` and still compile. It then fails at run time, because it is called without its object.

PascalThuet:
Thanks for the very fast and thorough follow-up.

From the dash.js use case, I wonder whether this RFC combines two separate steps: exposing a consumer-defined one-pass builder, and adding incremental input. dash.js currently passes complete strings to `DashParser`, while #432 already provides the shared scanner and internal builder.

Would it make sense to expose a small one-shot builder API first, and defer `write/end` and streaming semantics until a concrete consumer needs them?

The benchmark should also use #432 with an optimized `processNode` as its baseline, since the tuned result includes independent `Set.has` and `<S>` optimizations.

@dsilhavy, could you confirm whether a one-shot builder would cover the current dash.js need, and whether exact compatibility with the existing MPD, Patch and XLink output is required?

rfc/xml-incremental-parser.md
littlespex: Adopted in c66822a: `buildXml(input, builder, { root })`, and `createDocument` is optional. The prototype's `buildXml.ts` seeds the scanner's stack with the root, so the change is in the wrapper only, as you said. Tested under 'contract: root injection' in `prototype/equivalence.ts`.

rfc/xml-incremental-parser.md
littlespex: Adopted. `appendChild(parent, child, name)` receives the child's tag name as written. The text callbacks receive the parent's tag name. Both come from the `names` stack that the scanner already keeps. Implemented in `prototype/scan.ts` and documented under Callback semantics in the RFC. The flattening example in the guide handles plain `<S>` records by that name.

rfc/xml-incremental-parser.md
littlespex: Done. The Summary now says 'with no intermediate tree'. The guide has a new 'Flattening builders' section with a HAM-style example. It keeps a record per `<S>` and expands the records when the Representation or the Period ends. The section states the document-order dependency and the XSD sequence. Your adapter numbers (tree about 9 percent, URL templating about 60 percent) are in `findings.md` under consumers.

rfc/xml-incremental-parser.md
littlespex: Adopted both. Comments and doctypes are delivered as inner text, and the tree builder adds back `<!--`, `-->`, and `!` for `XmlNode`. Attribute names may start with any NameStartChar (`_`, `:`, non-ASCII letters). This is fixed in the shared scanner for both entry points, as a bug fix. `equivalence.md` lists the eight parity-corpus inputs whose output changes, with reasons. Two malformed-comment inputs change shape as a side effect (`<!--->` is reported as `<!---->`). The RFC says so.

rfc/xml-incremental-parser.md
littlespex: Aligned in c66822a. The RFC's type blocks use property syntax, as `XmlBuilder.ts` on `main` does, and every function-valued property has `this: void`, as you suggested. The prototype has a file-level disable for `no-invalid-void-type`. `steps.md` adds `allowAsThisParameter` to the repository config in the implementation.

rfc/xml-incremental-parser.md
littlespex: Adopted. `XmlParseError extends Error` with `offset`, `line`, and `column`. The message text is the same, and both entry points throw it (`prototype/XmlParseError.ts`). `keepWhitespace` stays an option. This RFC has no streams wrapper and no processing-instruction callback. Question 6 is settled in the text.

rfc/xml-incremental-parser.md
littlespex: Adopted under Future possibilities. A HAM builder is the second reference consumer, before any objectify builder.

plans/xml-incremental-parser/steps.md
littlespex: Rewritten against `main` after #432 (c66822a). The steps now port the prototype's scanner changes into `libs/xml/src/scan.ts`, add `buildXml` and the public types, regenerate exactly the eight changed fixture entries, and record the benchmark before and after. Incremental input is deferred to its own RFC, so the step to re-run the benchmark after the streaming guards is removed.

rfc/xml-incremental-parser.md
littlespex: Made it a real skip. `undefined` from `createElement` skips the subtree: no callbacks inside, no `appendChild` for it, and close tags are still checked. Measured with the check in the main loop: `parseXml` on the phase-one scanner is within -2 to +3 percent of `main` on all six inputs, so the check stays. Tested under 'contract: returning undefined skips the element' in `prototype/equivalence.ts`.

rfc/xml-incremental-parser.md
littlespex: Fixed in c66822a. The guide example is labeled illustrative, and it no longer claims to replace `processNode`. The faithful port of `processNode` and a faithful one-pass builder are in `prototype/dash.ts`. `prototype/equivalence.ts` asserts `deepStrictEqual` output on the synthetic manifests, `bbb_30fps.mpd`, the livesim2 fixture, and a namespaced sample with CDATA, mixed content, and repeated `Location` elements.

rfc/xml-incremental-parser.md
littlespex: Agreed. This is one reason incremental input moved to a follow-up RFC. 'Deferred: incremental input' in `findings.md` records the fix for that RFC: keep lexical progress across writes for text, comments, and CDATA, so that nothing is rescanned. Add a limit on the size of a buffered construct that throws when exceeded. Test with multi-megabyte constructs in small chunks.

rfc/xml-incremental-parser.md
littlespex: Adopted. Text is delivered untrimmed and entity-decoded. Whitespace-only runs are dropped unless `keepWhitespace` is set, which connects this to your note on question 4. `parseXml` trims in its tree builder. `appendCdata` falls back to `appendText` when absent. Both changes are in `prototype/scan.ts` and tested.

rfc/xml-incremental-parser.md
littlespex: Adopted for the single-call entry point. `buildXml` throws `Unexpected end of input inside ...` for an element left open or a construct cut off by the end of the input. The 22 truncated corpus inputs are asserted to throw. `parseXml` keeps its documented tolerance through an internal flag. The same rule is recorded for the streaming follow-up.

plans/xml-incremental-parser/benchmark.md
littlespex: Fixed in 2426f1d. The prototype is executable TypeScript under `plans/xml-incremental-parser/prototype/`. It runs against `main` after #432 with the checked-in livesim2 fixture (its sha256 is in `parseXml.equivalence.json` and quoted in `benchmark.md`). The repository's lint and typecheck cover it. `equivalence.ts` reports 778 checks, 0 failures, and 34 documented differences. The streaming JavaScript now exists only in history (8eb767e).

plans/xml-incremental-parser/benchmark.md
littlespex: Agreed. Those tables are withdrawn from the RFC, together with the streaming scope. `findings.md` records the method for the follow-up: flatten one whole input, then measure the flat whole input, decoder-like flat chunks, and `slice()` chunks, and report the slicing cost and the maximum duration of one write. The phase-one benchmark uses the generator on `main`, which returns a flat string.

plans/xml-incremental-parser/benchmark.md
littlespex: Withdrawn. `benchmark.md` now has two-builder pairs (`parseXml` and one custom builder alternating in one process), reported against the isolated medians. Each stays within about 6 percent on the large inputs. Randomized repeated processes and browser engines are not done, and the RFC says so.

rfc/xml-incremental-parser.md
littlespex: Done in c66822a. The baseline is `main` after #432. `parseXml` still calls `scan` directly, and the wrapper is `buildXml`. The result is stated plainly. The contract changes cost `parseXml` nothing measurable. A faithful one-pass replacement of `processNode` gains dash.js 1 to 9 percent on the stress inputs and nothing on the real ones. A lean builder gains 13 to 21 percent on the stress inputs and single digits on the real ones. Measured results and Drawbacks both say so.

rfc/xml-incremental-parser.md
littlespex: Fixed. The Summary example now keeps `attributes` in a nested object instead of spreading them.

plans/xml-incremental-parser/prototype.md
littlespex: Recorded under 'Deferred: incremental input' in `findings.md` for the follow-up RFC: the final flush belongs inside the same failure transition as scanning. Phase one has no `write`/`end` lifecycle. `buildXml` is a single call.

rfc/xml-incremental-parser.md
littlespex: Fixed in the shared scanner for both entry points, not behind a compatibility mode. Processing instructions end at `?>`. A `>` inside a quoted doctype literal no longer ends the doctype. `<?pi a > b?><root/>` and `<!DOCTYPE a SYSTEM "x>y"><a/>` are in the corpus as expected differences. `steps.md` lists them as Fixed entries for the changelog.

rfc/xml-incremental-parser.md
littlespex: Adopted. `XmlBuilder<TElement, TDocument = TElement>` with `createDocument(): TDocument`, `createElement(parent: TDocument | TElement, ...)`, `appendChild(parent, child: TElement, ...)`, and `buildXml` returning `TDocument`. It works with `{ root }` as you expected.

littlespex:
The RFC now follows the split you suggested (c66822a).

- Phase one is the builder contract plus `parseXmlWith(input, builder, options)`, which parses a complete string over the scanner from #432. `write`/`end` and incremental input move to their own RFC. The design notes stay in the design record.
- The benchmark baseline is now `main` after #432 (`parseXml` plus `processNode`). The tuned-builder rows are labeled as player-side specialization. On that baseline, a lean builder gains dash.js 13 to 21 percent on the stress manifests and single digits on the livesim2 manifest. A faithful one-pass replacement of `processNode` gains almost nothing. The RFC says so.
- The faithful builder in the prototype produces objects identical to today's pipeline. So exact compatibility with the MPD, Patch, and XLink output is possible if dash.js requires it. Your question to Daniel about whether it does still stands.

PascalThuet:
Thanks, this addresses my concerns. Separating incremental parsing from the one-shot builder and rebasing the benchmarks on #432 makes the scope much clearer.

The revised results also clarify the trade-off: a faithful builder brings almost no gain on real-sized manifests, while a lean dash.js-specific builder gains 13–21% on stress inputs and about 6% on the real livesim2 manifest.

The remaining question is whether dash.js can use the lean output or requires exact compatibility. I’ll leave that decision to @dsilhavy and the dash.js maintainers.

dsilhavy:
@PascalThuet @littlespex Thanks I like the one pass approach. 

> So exact compatibility with the MPD, Patch, and XLink output is possible if dash.js requires it

> The remaining question is whether dash.js can use the lean output or requires exact compatibility

What would be the implication of the lean output? For example for MPDPatching we need to be accurate and keep the order of remove, insert etc operations. My suggestion is to go ahead with exact compatibility as a first step and then improve as a next step if possible and after careful testing.

dsilhavy:
After discussion with @littlespex in the dash.js call today (comment purely from a dash.js side):

* The gain that we get from one-pass parsing seem to be negligible. Also for dash.js we have to walk up the tree a lot of times to get values from the parent elements
* We will get an immediate speed bump in dash.js by using the latest version of the XML parser
* Future improvements can come from a streaming parser which means we parse the XML as soon as we get any data.

Issue #112, opened 2024-10-28
Incorporate Standard Error Codes into CML


littlespex:
The Standard Error Code project aims to create a common set of error codes for media playback. Once the document is finalized, they should be incorporated into CML as:
- An enum or set of consts for the codes
- A set of mapping helper functions for converting player error events into standard error codes. Helpers should be made available for dash.js, hls.js and shaka-player.

PR #445, opened 2026-09-07
feat(error-codes): add @svta/cml-error-codes for SVTA2070 error codes

littlespex:
## Description

Implements [`rfc/error-codes.md`](https://github.com/streaming-video-technology-alliance/common-media-library/blob/main/rfc/error-codes.md), which merged in #423. `@svta/cml-error-codes` is the reference implementation of SVTA2070: Standardized Error Codes.

- `SvtaErrorCategory` and one code catalog per category, 137 named codes. Every code and category is also an individual `SVTA_*` constant, 146 in total, following the const enum pattern.
- `getSvtaErrorCategory` and `getSvtaErrorIndex` for the category and index arithmetic. `httpStatusToSvtaErrorCode` and `vastErrorToSvtaErrorCode` embed HTTP statuses and IAB VAST error codes.
- `vastErrorToSvtaErrorCode(1009)` returns 7999, per the review thread on #423. The RFC text now states this mapping, and its status is `accepted`.
- The first release is codes-only. `getSvtaErrorDescription`, the human-readable dictionary, is deferred until the IANA registry exists, per review of this PR. This resolves unresolved question 2 of the RFC. The member TSDoc keeps the spec description of every code.
- The package is wired into the root build script, the typedoc paths, and the root README list. It stays out of `scripts/projects.ts` until the first release-prep PR adds it to the publish list, so nothing publishes before that.

The code was written on `claude/svta-error-codes-cml-d0ad01` next to the RFC draft and ported here after the RFC merged. The port applies the writing style rules to the README, changelog, and TSDoc, and keeps the spec descriptions verbatim.

## Packages Changed

| @svta/cml-error-codes | 0.0.1 (new) | feat |

## Test Plan

- [x] `npm test` at the root: lint, build of every package, typecheck, and every package's tests pass
- [x] `npm test -w libs/error-codes`: 43 tests pass, including per-catalog invariants (unique integers, `floor(v / 1000)` equals the category, `UNKNOWN === category * 1000`, exact sizes, one `SVTA_*` constant per member), the codes the spec uses in its examples, and the VAST 1009 mapping
- [x] `npm run build -w docs`: 0 errors, no link warnings from this package
- [x] `libs/error-codes/config/cml-error-codes.api.md` reviewed against the RFC tables

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated

Refs: #112

https://claude.ai/code/session_01Mrjp8cavcuzBiQcBMRof6M


libs/error-codes/src/SvtaAdvertisingErrorCode.ts
```diff
@@ -0,0 +1,438 @@
...
+export const SVTA_VPAID_ERROR = 7901 as const
+
+/**
+ * SVTA 7 [Advertising] 999: VAST response document was empty
+ *
+ * SVTA2070 maps the four-digit IAB VAST error 1009 to this code.
+ *
+ * @public
+ */
+export const SVTA_EMPTY_VAST_RESPONSE = 7999 as const
```
littlespex, line 241: Should this map to `7009`?

littlespex: I kept 7999 for three reasons. The spec table is the authority, and I do not have the PDF in this environment, so please check it against the published row.

1. The transcribed advertising table (design doc, plan, and RFC) lists index 999 as `VAST response document was empty`, inside the 7100–7999 embed range.
2. Your review on #423 stated the same mapping: SVTA2070 requires the four-digit VAST error 1009 to translate to 7999.
3. 7009 is index 9. SVTA2070 reserves indices 0–99 for its own native codes and 100–999 for embedded external codes, so an external VAST code at 7009 would sit in the native range.

If the published table says 7009, say so and I will change the constant, the dictionary entry, the `vastErrorToSvtaErrorCode` special case, the tests, the README, and the RFC row in one push.

---

PR #452, opened 2026-09-09
refactor(cmcd): consolidate fan-out, dirty tracking, msd commit, and report copy

littlespex:
## Summary

Lands the four structural items the #422 review deferred, as one PR stacked on `refactor/cmcd-reporter-architecture`. The diff here shows only the five follow-up commits. Merge #422 first. GitHub then retargets this PR to `main`.

- One `runFanOut()` epilogue owns the fan-out depth, the queue drain, and the held eviction (693829a3). The three hand-copied recipes differed: the tick ran the drain inside a `finally` with no catch, so a drain throw replaced the original error and skipped `maybeEvict()`. `requestEvict()` collapsed into its one caller.
- The outbox is the single dirty channel (e60a8d4d). It reports through `onDirty` on push, on any synchronous exit that leaves lines queued, and on a failed send's re-queue. The facade never tracks queue state itself.
- Event mode commits the `msd` gate from the prepared report (f665be65). Request mode already used this rule, and the counters design needs it per target. `stampReport()` loses the `attachMsd` flag and its return value.
- `copyReportValues` is one recursive copy (ec0efda0). The split dispatch forced the `Date` fix on the parent branch to be written twice.

## Wire behavior

No wire change. One flag-only difference, disclosed in the changelog: a version 1 event target that enabled `msd` used to consume the per-target gate without `msd` ever reaching a version 1 wire. The gate now stays open. No report can observe the difference, because that target still never emits `msd`.

## Review threads

The four open #422 threads that requested these changes stay open until this PR lands, each with a reply pointing here: the fan-out helper, the dirty channel, the prepared-output `msd` commit, and the copy collapse.

## Verification

- Full root `npm test` passes: lint, build all, typecheck, and every package's tests, 18 suites with zero failures.
- `cml-cmcd.api.md` is byte-identical, so the public API is untouched.
- No test was added or edited. The parent branch's 563 cmcd tests pin the consolidated behavior and pass after every commit.


libs/cmcd/src/CmcdOutbox.ts
```diff
@@ -48,20 +49,29 @@ export class CmcdOutbox {
...
+				this.queue.unshift(...events)
+				this.onDirty()
+			})
 
-		this.send(events).catch(() => {
-			this.queue.unshift(...events)
+			dispatched = true
+		}
+
+		// One channel for "still holds lines": every synchronous exit that
```
littlespex: This wording is very unclear. What does this line mean?

libs/cmcd/src/CmcdOutbox.ts
```diff
@@ -48,20 +49,29 @@ export class CmcdOutbox {
...
+		// leaves the queue non-empty re-marks through onDirty, so no caller
+		// tracks queue state itself. The asynchronous failure path above is
+		// the one other marker, because its re-queue lands after this exit.
```
littlespex: The wording of this comment is confusing.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -416,17 +401,38 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
 
 	/**
-	 * Evicts retained sessions beyond the configured retention. An eviction
-	 * requested during a fan-out is held until the outermost fan-out has
-	 * drained, so a session an active fan-out still holds outlives the
-	 * request.
+	 * Runs one fan-out under the shared epilogue: the depth guard that defers
+	 * eviction, the queue drain, and the held eviction, in that order. A
+	 * throw from the fan-out or from the drain surfaces only after the whole
+	 * epilogue ran, and the first error wins.
```
littlespex: This comment is hard to follow. Reword it using the repo's writing style rules.

libs/cmcd/src/CmcdReporter.ts
```diff
@@ -521,20 +525,15 @@ export class CmcdReporter<C = Record<string, unknown>> {
...
-		// Surfaced only once every target has had its turn and the queues have
-		// been processed. Transforms must not throw; this makes the violation
-		// visible without letting it starve unrelated targets.
-		if (failure) {
-			throw failure.error
-		}
+			// Surfaced only once every target has had its turn, and after the
+			// fan-out epilogue has processed the queues. Transforms must not
+			// throw; this makes the violation visible without letting it
+			// starve unrelated targets.
```
littlespex: Reword this using the repo's writing style rules.

libs/cmcd/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to
 ### Changed
 
 - Internal restructuring of `CmcdReporter` into session-ledger, playback-state, report-pipeline, and outbox units in preparation for child reporters and automatic session counters. No public API change. Event-report processing now tracks only sessions with queued reports instead of scanning every retained session. Consumers that import `CmcdReporter` pay about 2.8 KB more minified (0.8 KB gzipped). Consumers of the codec functions alone are unaffected
+- Internal consolidation from the reporter architecture review. One fan-out epilogue owns the depth guard, the queue drain, and the held eviction. The outbox is the single source of dirty marks. The report copy is one recursive function. Event mode consumes the `msd` send gate from the prepared report, the same rule request mode uses. A version 1 event target that enabled `msd` no longer consumes the gate, which no version 1 report could ever carry
```
littlespex: This changelog entry, and the one above it, are too long. Rewrite this using the repo's writing style guidelines.

libs/cmcd/src/CmcdOutbox.ts
littlespex: Reworded in c048285a. The line meant: the queue still holds unsent lines, and onDirty reports that to the reporter. The comment now says that directly.

libs/cmcd/src/CmcdOutbox.ts
littlespex: Reworded in c048285a. The sentence explains why the failed-send callback keeps its own onDirty call: its re-queue runs after process() has returned.

libs/cmcd/src/CmcdReporter.ts
littlespex: Reworded in c048285a per the style rules.

libs/cmcd/src/CmcdReporter.ts
littlespex: Reworded in c048285a per the style rules.

libs/cmcd/CHANGELOG.md
littlespex: Both entries are rewritten in c048285a with shorter sentences and no internal jargon. The restructuring entry keeps the bundle cost and the API statement. The consolidation entry keeps the version 1 gate disclosure.

PR #453, opened 2026-09-09
fix(cmcd): honor the CTA-5004-B MUST NOT directives

littlespex:
## Summary

- The CMCD v2 implementation did not honor seven of the ten "MUST NOT" rules in CTA-5004-B. A probe against the built package found the gaps on the encoder, the reporter, and the validators.
- `prepareCmcdData` now omits `d` and `tpb` for other object types, omits `ab`, `lab`, and `tab` when the exact bitrate key is sent, omits `cen` unless the event is `ce`, and no longer turns `null` into `0` on the keys with a rounding formatter. Every encode path and `CmcdReporter` share this function.
- `CmcdReporter` event report bodies no longer end with a line feed.
- `validateCmcdStructure` reports the object type and aggregate bitrate conflicts as errors. `validateCmcdEvents` reports a trailing line feed as an error. Event reports from earlier `@svta/cml-cmcd` versions end with one, so a receiver that upgrades marks them invalid.

## Packages Changed

## Test Plan

- [x] `npm run build -w libs/cmcd` and `npm test -w libs/cmcd`: 588 tests pass, 24 of them new
- [x] `npm run typecheck` and `eslint .` pass
- [x] A probe of all ten directives against the built package passes on every surface. The `br: []` edge is left out on purpose.

Notes for review:

- The object type rules apply to version 2 payloads only. CTA-5004 (v1) has no object type restriction for `d`, so version 1 output is unchanged.
- The trailing line feed is an error in the validator. A warning was considered for compatibility with earlier clients and rejected in review.


libs/cmcd/src/prepareCmcdData.ts
```diff
@@ -180,8 +182,28 @@ export function prepareCmcdData(obj: Record<string, any>, options: CmcdEncodeOpt
 	for (const key of keys) {
 		let value = data[key] as CmcdValue
 
+		// An aggregate bitrate key is not sent alongside its exact bitrate key
+		const exactKey = CMCD_AGGREGATE_BITRATE_KEYS[key]
+		if (exactKey && keys.includes(exactKey) && isValid(data[exactKey])) {
+			continue
```
littlespex, line 196: **[P2] Preserve aggregate bitrate when formatting removes the exact bitrate**

The check reads the exact bitrate before formatting. With `{ab:[5000],br:[3000]}` and `formatters:{br:()=>NaN}`, encoding returns only `v=2`. The previous implementation retained `ab=(5000)`. The same problem affects `lab/lb` and `tab/tb`. Apply suppression after formatting confirms that the exact bitrate will be sent.

libs/cmcd/src/prepareCmcdData.ts
```diff
@@ -180,8 +182,28 @@ export function prepareCmcdData(obj: Record<string, any>, options: CmcdEncodeOpt
...
+		}
+
+		// Some keys are only sent for certain object types
+		const objectTypes = version > 1 ? CMCD_KEY_OBJECT_TYPES[key] : undefined
+		if (objectTypes) {
+			const ot = toTokenString(data['ot'])
+			if (ot !== undefined && !objectTypes.includes(ot)) {
+				continue
```
littlespex, line 209: **[P2] Treat an empty object type as unknown**

With `{ot:'',d:4000,tpb:[5000]}`, encoding now returns only `v=2`. An undefined object type preserves both metrics. The empty string triggers suppression before encoding omits it. This check also ignores custom formatting of `ot`, which can incorrectly discard metrics for a normalized video token. Evaluate the usable, formatted object type.

littlespex (commented):
Two P2 findings can silently discard metrics. Both findings were verified against the PR's base commit.

Build, typecheck, and lint pass. Tests report 589 passes and two TODOs. The API report is unchanged.

libs/cmcd/src/prepareCmcdData.ts
littlespex: Fixed in 7394e615. The aggregate check now evaluates the exact key after formatting, through the same `formatValue` helper the loop uses, so a formatter that empties `br` keeps `ab=(5000)`. Same for `lab`/`lb` and `tab`/`tb`. Test: `keeps the aggregate bitrate key when a formatter removes the exact key`.

libs/cmcd/src/prepareCmcdData.ts
littlespex: Fixed in 7394e615. The gate reads the formatted `ot` and treats an empty value as unknown, so `{ot:'', d, tpb}` keeps both keys, and a custom `ot` formatter is honored in both directions (`video` to `v` keeps `d`, `v` to `m` drops it). Per your follow-up, the gate still reads the data rather than the wire, so `d` stays out of a manifest request even when `ot` is not in `enabledKeys`. Tests: `treats an empty object type as unknown`, `reads the object type after formatting`, `drops d for a manifest when ot is filtered out`.

PR #454, opened 2026-09-09
fix(cmcd): flag requests with CMCD in both headers and query

littlespex:
## Summary

- `validateCmcdRequest` validated the CMCD headers when any were present and ignored a `CMCD` query parameter on the same request. CTA-5004 (v1) says a request MUST NOT carry both, and CTA-5004-B says each report MUST use only one transmission mode.
- The validator now reports an error when both are present. It still validates the headers and returns their data, so a receiver keeps the other findings.
- The query parameter is read from the query string instead of the full URL. A header-only request with a relative URL keeps working, and a query-only request with a relative URL no longer throws. A `CMCD` parameter inside the URL fragment is not a query parameter. When it is the only CMCD on the request, the error message says so, because a server never receives a fragment.

## Packages Changed

## Test Plan

- [x] `npm run build -w libs/cmcd` and `npm test -w libs/cmcd`: 555 tests pass, 10 of them new (`Request` and `HttpRequest` conflict cases, headers still validated, empty `CMCD=` not flagged, relative URLs, URL fragments)
- [x] `npm run typecheck` and `eslint .` pass

Notes for review:

- `CmcdReporter` never emits both modes by itself: `createRequestReport` is a `switch` on `transmissionMode`. It does copy the incoming headers, so a caller that already set `CMCD-*` headers and runs the reporter in query mode produces both. That is the caller's request, not the reporter's output.
- This branch and #453 both add a `### Fixed` block under `## [Unreleased]` in the cmcd changelog. Whichever merges second has a small changelog conflict.

Refs: #453

### DIFF
--- a/libs/cmcd/CHANGELOG.md
+++ b/libs/cmcd/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to
 
 ### Fixed
 
-- `validateCmcdRequest` reports an error when a request carries CMCD data in both the headers and the `CMCD` query parameter. CTA-5004-B allows one transmission mode per request, and CTA-5004 (version 1) says a request MUST NOT carry both. The headers are still validated and their data is returned. The query parameter is now read from the query string without parsing the full URL, so a relative URL no longer throws
+- `validateCmcdRequest` reports an error when a request carries CMCD data in both the headers and the `CMCD` query parameter. CTA-5004-B allows one transmission mode per request, and CTA-5004 (version 1) says a request MUST NOT carry both. The headers are still validated and their data is returned. The query parameter is now read from the query string without parsing the full URL, so a relative URL no longer throws. A `CMCD` parameter inside the URL fragment is not read as a query parameter. When the fragment is the only place that holds one, the error message says so, because a server never receives a fragment
 
 ## [2.6.1] - 2026-09-07
 
--- a/libs/cmcd/CHANGELOG.md
+++ b/libs/cmcd/CHANGELOG.md
@@ -10,7 +10,6 @@ and this project adheres to
 
 ### Fixed
 
-- `validateCmcdRequest` reports an error when a request carries CMCD data in both the headers and the `CMCD` query parameter. CTA-5004-B allows one transmission mode per request, and CTA-5004 (version 1) says a request MUST NOT carry both. The headers are still validated and their data is returned. The query parameter is now read from the query string without parsing the full URL, so a relative URL no longer throws. A `CMCD` parameter inside the URL fragment is not read as a query parameter. When the fragment is the only place that holds one, the error message says so, because a server never receives a fragment
 
 ## [2.6.1] - 2026-09-07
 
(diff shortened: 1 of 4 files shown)


libs/cmcd/src/validateCmcdRequest.ts
```diff
@@ -72,6 +89,18 @@ export function validateCmcdRequest(request: Request | HttpRequest, options?: Om
...
 }
 
+function getCmcdQueryParam(url: string): string | null {
+	const start = url.indexOf('?')
+
+	if (start < 0) {
+		return null
+	}
+
+	const end = url.indexOf('#', start)
```
littlespex: **[P2] Ignore question marks inside URL fragments**

The parser searches for `#` only after `?`. For `https://cdn.example.com/seg.mp4#t=10?CMCD=br%3D5000`, it treats fragment content as query data.

- With valid CMCD headers, validation incorrectly reports a transmission-mode conflict.
- Without headers, validation incorrectly accepts the fragment as CMCD data.

I reproduced both outcomes. Check the first fragment delimiter before extracting the query, and add regression tests.

littlespex (commented):
I found one issue to fix before merging, detailed in the inline comment.

The intended conflict check agrees with CTA-5004 v1 section 2. Preserving header data follows section 4, item 8.

Reviewed commit `7f28f1ee`. Build, full typecheck, and focused lint passed. CMCD tests: **551 passed**, two TODOs. The API report is unchanged.

libs/cmcd/src/validateCmcdRequest.ts
littlespex: Fixed in cc081ffb. `getCmcdQueryParam` now cuts the URL at the first `#` before it looks for `?`, so `#t=10?CMCD=br%3D5000` is fragment text. With CMCD headers the request validates clean. Without headers it reports no CMCD data, and per the follow-up discussion the message now says the fragment holds a `CMCD` parameter that a server never receives. Tests: the `URL fragments` block, four cases including a `Request` object.

libs/cmcd/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- `validateCmcdRequest` reports an error when a request carries CMCD data in both the headers and the `CMCD` query parameter. CTA-5004-B allows one transmission mode per request, and CTA-5004 (version 1) says a request MUST NOT carry both. The headers are still validated and their data is returned. The query parameter is now read from the query string without parsing the full URL, so a relative URL no longer throws. A `CMCD` parameter inside the URL fragment is not read as a query parameter. When the fragment is the only place that holds one, the error message says so, because a server never receives a fragment
```
littlespex: This entry is too long. Rewrite it using the repo's writing style guidelines.

littlespex: Rewritten in 010c3da3: two bullets, one per change, every sentence under 20 words, and the spec citation and rationale that the PR body already carries are dropped.

littlespex: Correction: 010c3da3 removed the old entry but did not insert the rewritten bullets, so the Fixed section was empty. a03abe20 adds the two rewritten bullets.

PR #455, opened 2026-09-10
[RFC] Session API for CMCD version 2

littlespex:
## Description

Draft RFC for a second reporting API in `@svta/cml-cmcd`, next to `CmcdReporter`, plus its design record. Review happens in the comments of this PR, per `rfc/README.md`.

**What the RFC proposes.** Three objects that match the three scopes of CTA-5004-B. A `CmcdSession` is one `sid` and owns the targets, the transport, the timers, and the session totals. A `CmcdPlayer` is one media player and pushes state, records events, decorates requests, and records responses. Targets are internal and own the per-destination state: `sn`, the `msd` gate, the `bs` flag, the `ec` buffer, and the queue. The reporter derives every key the spec defines in terms of state it can observe: `msd`, `bs`, `su`, `dl`, `h`, `bg`, the starvation counters, and the response keys. State-change events come only from `update()`. A session holds several players, and each interval tick emits one `t` line per live player. Late responses report under the session that issued the request through the record on the decorated request, with no retention window and no provenance symbol.

**Why now.** PR #422 restructures `CmcdReporter` while keeping every 2.6 semantic. Before committing to that path, this RFC asks which of those semantics a design built for the version 2 spec would keep. The comparison document answers that for each concept. Two findings from the hls.js and dash.js integrations, read on 2026-09-09, are in the Motivation section. Both players would drop every `rr` event on 2.6 or later, because neither passes the decorated request back. dash.js pairs `update({ sta })` with `recordEvent(PLAY_STATE, data)`, which loses the payload on 2.4 or later.

**Files.**

- `rfc/cmcd-session-api.md`: the proposal for adopters, with the API, one complete example, the contract, the costs, the alternatives, and nine open questions.
- `plans/cmcd-session-api/architecture.md`: the internal units, state, algorithms, the key table that replaces the spread of encoder tables, and the test plan.
- `plans/cmcd-session-api/comparison.md`: `CmcdReporter`, PR #422, and this design compared on size, concepts, spec coverage, and adopter impact.

**For reviewers.** The open questions in the RFC are the decisions this review should make, in particular the configuration shape, `bg` from document visibility by default, the `dl` default, and one `t` line per player during interstitials. `CmcdReporter` does not change in this proposal. Three spec gaps found on the way are listed at the end of the architecture document and can ship as separate fixes.

## Requirements Checklist

- [ ] Unit Tests updated or fixed (not applicable, no code changes)
- [ ] Docs/guides updated (not applicable, the RFC and the design record are the documents)
- [ ] Changelog updated (not applicable, no package change)

### DIFF
--- a/rfc/cmcd-session-api.md
+++ b/rfc/cmcd-session-api.md
@@ -103,6 +103,33 @@ const ad = session.createPlayer({ cid: 'ad-7' })
 
 Every player reports under the session's `sid` and takes the next sequence number of each target. Dispose a player when its media player is destroyed.
 
+The ownership in one picture. Targets belong to the session, and every player reports to every target.
+
+```mermaid
+flowchart TB
+    subgraph session["CmcdSession, one sid"]
+        direction TB
+        sdata["sid, version, transport, timers<br>msd, bg, bsa, bsda, completed spans"]
+        subgraph players["Players, one per media player"]
+            direction LR
+            primary["CmcdPlayer, cid movie-42<br>store, reported values, host, su, open span"]
+            ad["CmcdPlayer, cid ad-7<br>store, reported values, host, su, open span"]
+        end
+        subgraph targets["Targets, one per destination"]
+            direction LR
+            req["request target<br>sn, msd gate, bsd cursor<br>bs and ec per player"]
+            ev1["event target A<br>sn, msd gate, bsd cursor<br>bs and ec per player<br>queue, back-off, gone"]
+            ev2["event target B<br>same state as A"]
+        end
+    end
+    primary --> req
+    primary --> ev1
+    primary --> ev2
+    ad --> req
+    ad --> ev1
+    ad --> ev2
+```
+
 ### Push state, get events
 
 `update()` stores playback state and derives the five state-change events from it: `sta` gives `ps`, `pr` gives `pr`, `cid` gives `c`, `bg` gives `b`, and `br` gives `bc`. The event fires when the value differs from the last value the reporter reported for that player. Metrics pushed in the same call ride the event.
@@ -160,6 +187,26 @@ primary.recordResponse(req, {
 
 A response can arrive after the player has disposed the session that issued the request. The record on the request points to that session, so the `rr` report has the old `sid` and the old session's next sequence number. The old session sends the report at once, because no batch will fill again.
 
+The sequence for a response that arrives after its session was disposed:
+
+```mermaid
+sequenceDiagram
+    participant Player as media player
+    participant A as CmcdSession A
+    participant B as CmcdSession B
+    participant C as collector
+    Player->>A: player.decorate(request, data)
+    A-->>Player: request with cmcd record, sid A, sn 41
+    Player->>Player: send the request to the CDN
+    Player->>A: session.dispose()
+    A->>C: POST the queued lines
+    Player->>B: createCmcdSession(), then createPlayer()
+    Note over Player,C: the response for the old request arrives
+    Player->>B: player.recordResponse(request, info)
+    B-->>A: the record resolves to session A
+    A->>C: POST e=rr with sid A and sn 42, at once
+```
+
 ### Interstitials
 
 The spec covers multi-player sessions with `cid`, `ps`, and `nr`. The session API needs no other concept. When an interstitial renders on its own player, the primary player sets `nr` while it fetches content that the user does not see.
@@ -174,6 +221,23 @@ primary.update({ nr: false })
 ```
 
 Each interval tick emits one `t` line per live player, with the same `ts` and consecutive sequence numbers. Overlay and side-by-side interstitials report both players, and a collector tells them apart by `cid` and `nr`.
+
+The wire during a sequential interstitial. Keys other than the ones shown are omitted.
+
+```text
+# The primary pauses and stops rendering while the ad player starts
+cid="movie-42",e=ps,nr,sid="s1",sn=7,sta=a,ts=1764752400000,v=2
+cid="ad-7",e=as,sid="s1",sn=8,ts=1764752400050,v=2
+cid="ad-7",e=ps,sid="s1",sn=9,sta=p,ts=1764752400120,v=2
+
+# One interval tick: one t line per live player, same ts, consecutive sn
+cid="movie-42",e=t,nr,sid="s1",sn=10,sta=a,ts=1764752430000,v=2
+cid="ad-7",e=t,sid="s1",sn=11,sta=p,ts=1764752430000,v=2
+
+# The ad ends and its player is disposed, the primary renders again
+cid="ad-7",e=ae,sid="s1",sn=12,ts=1764752445000,v=2
+cid="movie-42",e=ps,sid="s1",sn=13,sta=p,ts=1764752445200,v=2
+```
 
 ### Lifecycle
 
@@ -319,6 +383,30 @@ The `cid` given to `createPlayer()` counts as reported, so creating a player emi
 
 `bsa` counts the transitions into `r`. `bsda` and `bsd` count completed spans only. A span still open at `dispose()` is dropped. Automatic `bsa`, `bsd`, and `bsda` entries have no cause token. `url` is the request URL without its `CMCD` parameter. `rc` is `0` when `status` is absent. `ts` for a response is the request start.
 
+The `sta` transitions and what each one derives. Transitions with no effect on a derived key are left out.
+
+```mermaid
+stateDiagram-v2
+    direction LR
+    d : d preloading
+    s : s starting
+    p : p playing
+    k : k seeking
+    r : r rebuffering
+    a : a paused
+    [*] --> d
+    [*] --> s : records msdStart, su true
+    d --> s : records msdStart, su true
+    s --> p : msd from msdStart, su false
+    p --> k : su true
+    k --> p : su false
+    p --> r : bsa increments, span opens, bs set on every target, su true
+    r --> p : span closes into bsda and bsd, su false
+    r --> a : span closes into bsda and bsd
+    p --> a
+    a --> p
+```
+
 ### Reports and targets
 
 Every report is assembled in a fixed order, and a later source wins:
@@ -330,6 +418,25 @@ Every report is assembled in a fixed order, and a later source wins:
 5. the derived defaults
 
 The target's `transform` runs on a copy when one is configured, and `null` cancels the report for that target. The reporter then filters the keys, applies the spec rules, encodes, and only then commits: `sn` advances, the `msd` gate closes when the output kept `msd`, `bs` and the `ec` buffer clear, and the `bsd` cursor moves. A cancelled or failed report commits nothing.
+
+The report path for one target:
+
+```mermaid
+flowchart LR
+    subgraph assemble["Assemble, a later source wins"]
+        direction TB
+        s1["player store"] --> s2["session data"] --> s3["per-call data"] --> s4["target state for the player"] --> s5["derived defaults"]
+    end
+    assemble --> tq{"transform<br>configured?"}
+    tq -- no --> prep["filter keys<br>apply the spec rules"]
+    tq -- yes --> tr["copy nested values<br>run the transform"]
+    tr -- null --> cancel["cancelled<br>nothing committed"]
+    tr -- data --> restore["restore required keys<br>re-stamp sid, e, ts"] --> prep
+    prep --> enc["encode"]
+    enc -- throws --> fail["thrown to the caller<br>nothing committed"]
+    enc -- line --> commit["commit<br>sn advances<br>msd gate closes if kept<br>bs and ec clear<br>bsd cursor moves"]
+    commit --> out["event target: queue the line<br>request target: URL or headers"]
+```
 
 Filtering follows the target's `keys`. A key the current event requires is included whatever the list says:
 
@@ -408,8 +515,26 @@ Each event target queues encoded lines. It sends a batch when the queue reaches 
 | 2xx | done, back-off resets |
 | 410 | the target sends nothing else for this session |
 | 429, 5xx, or a rejected transport | the batch returns to the front of the queue, and the target retries after a back-off |
+| other 4xx | the batch is dropped, back-off resets |
 
 The back-off starts at one second and doubles to a cap of 60 seconds. New lines keep queueing during the back-off and go out with the retry, which is the aggregation the spec recommends for 429. After `dispose()`, a target stops retrying when a retry at the 60 second step fails. When a queue exceeds `maxQueueSize`, the oldest lines are dropped.
+
+The delivery states of one event target:
+
+```mermaid
+stateDiagram-v2
+    [*] --> queueing
+    queueing --> sending : batchSize reached, flush, dispose, or a late rr after dispose
+    sending --> queueing : 2xx, back-off resets
+    sending --> queueing : other 4xx, batch dropped
+    sending --> gone : 410
+    sending --> backingOff : 429, 5xx, or a rejected transport, batch back to the front
+    backingOff --> sending : timer fires, 1 s doubling to 60 s, or flush
+    backingOff --> gone : after dispose, the 60 s retry fails
+    gone --> [*]
+    note right of queueing : lines past maxQueueSize drop from the front
+    note right of backingOff : new lines keep queueing and join the retry
+```
 
 ### Errors
 
--- a/rfc/cmcd-session-api.md
+++ b/rfc/cmcd-session-api.md
@@ -1,0 +1,521 @@
+---
+status: draft
+---
+
+# RFC: Session API for CMCD version 2
+
+| | |
+|---|---|
+| **Author** | Casey Occhialini |
+| **Date** | 2026-09-09 |
+| **Package** | `@svta/cml-cmcd` |
+| **Breaking change** | No |
+
+## Summary
+
+Add a second reporting API to `@svta/cml-cmcd`, next to `CmcdReporter`. The new API has three objects that match the three scopes of CTA-5004-B.
+
+- A `CmcdSession` is one `sid`. It owns the report targets, the transport, the interval timers, and the session totals.
+- A `CmcdPlayer` is one media player inside the session. A player pushes its state, records events, decorates its requests, and records their responses.
+- A target is one report destination. Targets are internal. Each target owns the state the spec scopes to a destination. That state is the sequence number, the `msd` gate, the `bs` flag, the `ec` buffer, and the delivery queue.
+
+The reporter derives every key the spec defines in terms of state the reporter can observe. A player never computes `msd`, `bs`, `su`, `sn`, the starvation counters, or the response timing keys. State-change events come only from `update()`. A session can hold several players, and the wire shows one `sid` and one sequence per target.
+
+```ts
+import { createCmcdSession, CmcdEventType } from '@svta/cml-cmcd'
+
+const session = createCmcdSession({
+	keys: ['br', 'bl', 'd', 'ot', 'sid', 'cid', 'mtp', 'sf', 'st', 'su', 'nor'],
+	eventTargets: [{ url: 'https://collector.example.com/cmcd' }],
+})
+
+const player = session.createPlayer({ cid: 'movie-42' })
+
+player.update({ sf: 'h', st: 'v' })
+player.update({ sta: 'p', bl: 3200, mtp: 15000, pt: 12000 }) // emits e=ps with this snapshot
+
+const req = player.decorate({ url: 'https://cdn.example.com/seg-1.m4s' }, { ot: 'v', d: 4000, br: 3000 })
+const res = await fetch(req.url, { headers: req.headers })
+player.recordResponse(req, { status: res.status, headers: res.headers }) // emits e=rr
+
+player.recordError('MEDIA_ERR_NETWORK')
+player.recordEvent(CmcdEventType.AD_BREAK_START)
+
+session.dispose()
+```
+
+`CmcdReporter` does not change. It remains until players have migrated, and a later release marks it deprecated.
+
+## Motivation
+
+CMCD version 2 is much larger than version 1. It adds a second reporting mode, several targets with their own filters, 19 event types, and 32 new keys. It also adds rules that scope state to a session or to a destination. `CmcdReporter` encodes and sequences reports, but it leaves much of the spec to the player. The two largest adopters show the result. The table lists spec behavior that hls.js and dash.js implement by hand, read from their development branches on 2026-09-09.
+
+| Spec behavior | hls.js | dash.js |
+|---|---|---|
+| `msd` from the `sta` starting to playing delta | not sent | `calculateMsd()` in `CmcdModel` |
+| `bs` per destination since the last report | `starved` flag, cleared on the next video request | `wasPlaying()` and `onRebufferingStarted()` |
+| `su` until the buffer is stable | `buffering` flag | set per request in `CmcdModel` |
+| `dl` from buffer length and playback rate | computed per request | computed per request |
+| `bsd` from rebuffering spans | not sent | `_rebufferingDuration` in `CmcdModel` |
+| `ec` buffered per destination | fatal errors only, one event | `update({ ec })`, so the code stays on every later report |
+| `cmsds` and `cmsdd` from response headers | not sent | base64 encoding in `CmcdController` |
+
+The current API also has three traps that the two integrations fall into.
+
+1. `update({ sta })` fires a `ps` event, and a following `recordEvent(PLAY_STATE, data)` is deduplicated. dash.js makes exactly that pair of calls. On `@svta/cml-cmcd` 2.4 or later its play-state events lose the `data` payload.
+2. `recordResponseReceived()` attributes a response only through the provenance record on the decorated request. The hls.js loader passes `{ url }` from its loader context. The dash.js interceptor copies only the `cmcd` object from the decorated request. On 2.6 or later both players drop every `rr` event. Both pin older versions today.
+3. Configuration is fixed at construction. Both players rebuild the reporter to apply a manifest setting, and the rebuild resets `sid` and every `sn`. dash.js has a comment about that reset.
+
+Multi-player sessions are the case the current design fits worst. The spec expects one `sid` to span a movie and its interstitials, and the `bg` key is defined over "all players in a session". A second `CmcdReporter` with the same `sid` restarts every sequence number and sends `msd` again. The child-reporters proposal ([PR #398](https://github.com/streaming-video-technology-alliance/common-media-library/pull/398)) adds a root and child split to the current class. That proposal calls a session-first API "the right long-term shape" and set it aside for surface area.
+
+Inside the package, `CmcdReporter` is one class with five jobs: session bookkeeping, report assembly, transform policy, delivery, and the public facade. Each feature since version 2.4 added pairwise interactions between those jobs. A refactor ([PR #422](https://github.com/streaming-video-technology-alliance/common-media-library/pull/422)) splits the class along those jobs while keeping every current semantic. This RFC asks a different question: which of those semantics would a design built for the version 2 spec keep at all? The answer removes session rotation, the retention window, eviction, the provenance record, and the `customData` generics, and adds derived keys and a player object. The design record in [`plans/cmcd-session-api/`](../plans/cmcd-session-api/) has the comparison.
+
+## Guide-level explanation
+
+### One session, one or more players
+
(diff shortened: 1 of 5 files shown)


rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+```ts
+const req = primary.decorate(
+	{ url: frag.url, headers: {} },
+	{ ot: 'v', d: 4000, br: 3000, nor: next.url },
+)
+// req.url has the CMCD query parameter, or req.headers has the four CMCD headers
+// req.cmcd.data is the report as sent, for logging or a player event
+```
+
+Hand the returned request back with the response. The reporter derives `url`, `rc`, `ts`, `ttfb`, and `ttlb`, and reads `cmsds` and `cmsdd` from the `CMSD-Static` and `CMSD-Dynamic` response headers.
```
littlespex, line 177: The provenance concept #422 exists to handle multiple session ids per playback session. Is there any guidance on how to handle this with the session APIs? It seems like another piece of machinery that each player would have to implement and maintain.

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+
+```ts
+import { createCmcdSession, CmcdEventType } from '@svta/cml-cmcd'
+
+const session = createCmcdSession({
+	keys: ['br', 'bl', 'd', 'ot', 'sid', 'cid', 'mtp', 'sf', 'st', 'su', 'nor'],
+	eventTargets: [{ url: 'https://collector.example.com/cmcd' }],
+})
+
+const player = session.createPlayer({ cid: 'movie-42' })
```
littlespex: The name `createPlayer` might be confusing for a player developer. What is being created is really a reporter. I understand that the "reporter" name also conflicts with things like `CmcdReporter`, but calling it a player doesn't seem right.

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+```ts
+type CmcdSessionConfig = {
+	sid?: string                                // default: a new UUID
+	version?: CmcdVersion                       // request mode only. default CMCD_V2
+	transmissionMode?: CmcdTransmissionMode     // default CMCD_QUERY
+	keys?: readonly CmcdKey[]                   // request mode allowlist. default: every key of the version
+	headerMap?: Partial<CmcdHeaderMap>          // header shard per custom key
+	transform?: CmcdRequestTransform
+	eventTargets?: readonly CmcdEventTargetConfig[]
+	transport?: CmcdTransport                   // default: fetch, POST, keepalive
```
littlespex: What is `CmcdTransport` exactly?

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+
+The dash.js change deletes `calculateMsd()`, the rebuffer tracking, the `ec` persistence, and the rebuild logic. It passes the decorated request through its interceptors.
+
+## Drawbacks
+
+- **Two reporting APIs in one package.** Until `CmcdReporter` is removed, both need documentation and tests. Adopters must choose.
+- **Two objects in the simplest integration.** A request-only player still creates a session and a player.
+- **One `t` line per player changes the wire during interstitials.** A collector that expected one interval line per `sid` sees two while an interstitial player exists.
+- **A DOM side effect.** The session listens to `visibilitychange` when a document exists. The `visibility` option turns that off.
+- **Derived defaults are assumptions.** `dl` from `bl` and `pr`, and `su` from the play state, match what hls.js and dash.js compute today. The spec words `dl` as a possible equivalence only.
```
littlespex, line 661: Could derived defaults be a configurable option?

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+- **Exact attribution needs the returned request.** A player that keeps only the URL gets current-session attribution for late responses.
+- **In-flight requests keep their session reachable.** There is no retention knob. Memory is bounded by the requests the player keeps.
+- **Event-mode code is always bundled** with the session API, because targets are configuration objects.
```
littlespex: Is that not also the case for the current `CmcdReporter` implementaion?

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+- **Shaka Player** `CmcdManager` is an independent implementation with the same per-player fields and no session sharing.
+- **The report transforms RFC** ([`rfc/cmcd-reporter-middleware.md`](./cmcd-reporter-middleware.md)) defined the transform contract this API reuses.
+- **The session retention RFC** ([`rfc/cmcd-session-retention.md`](./cmcd-session-retention.md)) defined the late-response goal this API meets by object lifetime.
+- **The child-reporters RFC** ([PR #398](https://github.com/streaming-video-technology-alliance/common-media-library/pull/398)) defined the state partition between a session and a player that this API makes the primary model.
+- **The starvation counters design** ([`plans/cmcd-session-counters/design.md`](../plans/cmcd-session-counters/design.md)) settled the supplied-value precedence this API uses for every derived key.
+
+## Unresolved questions
+
+1. **Configuration shape.** Top-level request settings plus `eventTargets`, as proposed, or a `targets` list of factory functions that tree-shakes event mode.
+2. **`bg` from document visibility by default.** On by default with an opt-out, or off by default.
```
littlespex: On by default with opt out

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,521 @@
...
+3. **`dl` as a derived default.** Keep it, or make `dl` player-supplied only.
```
littlespex: Does "derived default" mean it can still be overridden via an `update` call if need be?

rfc/cmcd-session-api.md
```diff
@@ -0,0 +1,646 @@
...
+    participant C as collector
+    Player->>A: player.decorate(request, data)
+    A-->>Player: request with cmcd record, sid A, sn 41
+    Player->>Player: send the request to the CDN
+    Player->>A: session.dispose()
+    A->>C: POST the queued lines
+    Player->>B: createCmcdSession(), then createPlayer()
+    Note over Player,C: the response for the old request arrives
+    Player->>B: player.recordResponse(request, info)
+    B-->>A: the record resolves to session A
```
littlespex: How would this work exactly? Do sessions have knowledge of each other? Is there a session tracking system at a different level?

rfc/cmcd-session-api.md
littlespex: Renamed in e8ef5fac. `session.createReporter({ cid })` returns a `CmcdSessionReporter`. The config type is `CmcdSessionReporterConfig`, and the payload type is `CmcdPlaybackData`. The prose now uses "player" only for the media player.

rfc/cmcd-session-api.md
littlespex: Defined in e8ef5fac, in the Configuration section. `CmcdTransport` is `(request: HttpRequest) => Promise<{ status: number }>`, the function that sends the event-mode POST requests. It is the `requester` argument of `CmcdReporter` under a new name, so the hls.js and dash.js loader adapters move over unchanged. The default is `fetch` with `keepalive`.

rfc/cmcd-session-api.md
littlespex: Added in e8ef5fac as `derive?: Partial<Record<'bg' | 'dl' | 'su', boolean>>`, all on by default. It replaces the separate `visibility` option. I chose the record over an array because the option is an opt-out. A player names the default it turns off. A partial array would read as an allowlist, where listing one entry silently disables the other two. The other derived keys are spec definitions and keep only the supply-wins rule.

rfc/cmcd-session-api.md
littlespex: Correct. The class has the same property. My baseline measured a request-only import of `CmcdReporter` at the same 18.8 KB as the full class. Reworded in e8ef5fac: the drawback now says it matches today, and that the factory-function alternative would improve on both.

rfc/cmcd-session-api.md
littlespex: Recorded in e8ef5fac as the resolution of open question 2. On by default, and `derive: { bg: false }` turns it off.

rfc/cmcd-session-api.md
littlespex: Yes. A derived default fills a key only when the player has not supplied it. A value from `update()` wins for the rest of the session, and a value in per-call data wins for that report. e8ef5fac defines the term at the start of the Derived keys section and records the resolution of question 3: `dl` stays a derived default, and `derive: { dl: false }` turns it off.

rfc/cmcd-session-api.md
littlespex: No session tracking exists. e8ef5fac adds the mechanism to the Requests section. The `cmcd` record on the returned request is the key in a reporter-internal `WeakMap`, and the value is the origin: the issuing session and reporter, the per-request data, and the start time. `recordResponse()` on any reporter looks the record up and reports through that origin, so a response routes to the right session whichever reporter receives it. The entry lives as long as the request object. Whether that removes the need for `sid` change guidance continues in the line 176 thread and in chat.

rfc/cmcd-session-api.md
littlespex: Adopted in d7bbdd72. `session.rotate(sid?)` starts the next `sid` for the whole session with no new objects. Every reporter keeps its store and `cid`, every target restarts `sn` at zero, the `msd` gate re-arms, the session totals reset, and the dedup baselines reset so the next push emits under the new `sid`. Rotation itself emits nothing, and the same `sid` is a no-op.

Late responses need nothing from the player either way. The request record points at the internal state of the `sid` that issued it, so `recordResponse()` on whichever reporter the player holds reports under the old `sid` with that `sid`'s next `sn`. The guidance is one sentence in the new "Changing the `sid`" section: `rotate()` for a new `sid` on the same playback, a new session for a new playback.

littlespex:
**Open questions resolved, 2026-09-10.** The Unresolved questions section of the RFC now records a resolution for every item. Commit 7ec32e4c has the text. Summary:

1. **Configuration shape.** Kept: top-level request settings plus `eventTargets`. The bundle gain of target factories is unmeasured, and no known adopter has a request-only configuration path. A request-only entry point is listed under Future possibilities.
2. **`bg` from document visibility.** On by default. `derive: { bg: false }` turns it off. Resolved 2026-09-09.
3. **`dl` as a derived default.** Kept. A supplied `dl` wins, and `derive: { dl: false }` turns the default off. Resolved 2026-09-09.
4. **One `t` line per live reporter.** The design stands. Spec section 4 describes a multi-player session as per-player reports, told apart by `cid`, play state, and `nr`. Section 8.1.7 shows two players under one `sid`. Collector maintainers can reopen this question here.
5. **`b` on exit.** A bare `e=b`, with no `bg` key. The `bg` key row says the key SHOULD only be sent when TRUE. The `b` event definition describes exit as the event without `bg`. A per-target transform is the opt-in for `bg=?0`, shown in the Transforms section. Side finding: `validateCmcdStructure` rejects a `b` event without `bg` today. A separate fix is in progress.
6. **`ts` placement.** Kept in the payload. Every reporting call takes one data shape.
7. **URL-only `recordResponse()`.** The current-session fallback only. Wire parsing would need a map of past `sid` states with an eviction rule, which is the ledger this design removes.
8. **`onError` signature.** A single callback. The error message names the target URL and the stage, one of transform, encode, or send. `cause` holds the requester's error.
9. **Names.** `CmcdSessionReporter` and `CmcdPlaybackData` accepted. The send function is renamed from `transport` and `CmcdTransport` to `requester` and `CmcdRequester`. The package already uses "transport" for the `CmcdReportRecorder` adapters, and `requester` is the existing `CmcdReporter` argument name.

Also fixed in the same commit: the derived keys table said a supplied `ts` is ignored. It wins for that report.

littlespex (commented):
Recommendation: request changes before accepting this RFC.

Reviewed [PR #455](https://github.com/streaming-video-technology-alliance/common-media-library/pull/455) at commit 7ec32e4cde7d609599de848aee8923e8d000214c. The head remained unchanged during verification.

I read all three changed documents and every returned PR discussion entry. The discussion contained the following:

- 16 inline comments, including replies and outdated comment contexts.
- 16 review submissions.
- One conversation comment resolving the open questions.

I also checked the accepted retention and transform RFCs, relevant earlier reviews, and current dash.js configuration behavior.

The findings below describe defects or missing rules in the proposed contracts. This PR contains no implementation to test. Only the TypeScript declaration failure was reproduced with the compiler.

Each finding includes an anchor in the reviewed revision. P1 findings affect central design guarantees. P2 findings should be resolved before implementation relies on the contract.

1. **[P1] Preserve request state per destination**

   [plans/cmcd-session-api/architecture.md:68](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L68)

   The design has one request target for the entire session. Requests to different CDNs therefore share its sequence, starvation flags, duration cursor, and error buffers.

   After a completed stall, a request to destination A consumes its pending starvation data. A later request to destination B receives none. Reports sent to A also advance B's sequence.

   Give request destinations explicit identities and separate target state. Keep an optional default identity for simple integrations. Do not infer identity solely from hostname.

   This requirement was already [accepted during the child-reporters review](https://github.com/streaming-video-technology-alliance/common-media-library/pull/398#discussion_r3755493650). That decision explicitly rejected the shared counter and explained why hostname cannot identify a destination. [CMCD defines destination-specific reporting state](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#sequence-number).

2. **[P1] Preserve immutable data for late responses**

   [rfc/cmcd-session-api.md:529](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L529)

   The response path reads the origin reporter's live store after rotation. The origin also retains raw per-request data without specifying a copy.

   For example, issue a request under A, rotate to B, and change stored metrics. The response retains A's identifier but reads B's updated metrics. Reusing and mutating the original per-request object can also change the reported object type or bitrate.

   Archive the reporter's store when its sid ends. Capture a detached copy of per-request values before running transforms. Nested metric values also need isolation. A WeakMap controls lifetime but does not provide snapshot semantics.

   The [accepted retention RFC](https://github.com/streaming-video-technology-alliance/common-media-library/blob/2653a012a3e98d0bd2d22d0269a75e828551dda3/rfc/cmcd-session-retention.md#L97) explicitly preserves archived snapshots. Its [per-request snapshot contract](https://github.com/streaming-video-technology-alliance/common-media-library/blob/2653a012a3e98d0bd2d22d0269a75e828551dda3/rfc/cmcd-session-retention.md#L212) explains the mutation failures this prevents.

3. **[P2] Define configuration changes before removing dash.js rebuilds**

   [rfc/cmcd-session-api.md:630](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L630)

   The migration says dash.js deletes its rebuild logic. However, rotation changes only sid, and session configuration remains fixed.

   Manifest parameters can change the request version, transmission mode, and key allowlist after construction. Removing the rebuild leaves the previous configuration active. Recreating the session still resets the counters, which preserves the problem identified in Motivation.

   Define configuration changes that preserve the appropriate session state. Alternatively, retain the rebuild requirement and narrow the migration claim.

   Verified upstream: [manifest processing marks the reporter for rebuilding](https://github.com/Dash-Industry-Forum/dash.js/blob/development/src/streaming/controllers/CmcdController.js#L465). The configuration gives manifest values priority for [mode](https://github.com/Dash-Industry-Forum/dash.js/blob/development/src/streaming/cmcd/config/CmcdPropertyMap.js#L150) and [keys](https://github.com/Dash-Industry-Forum/dash.js/blob/development/src/streaming/cmcd/config/CmcdPropertyMap.js#L206).

4. **[P2] Keep bs active throughout an ongoing stall**

   [rfc/cmcd-session-api.md:410](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L410)

   The algorithm sets bs only when entering rebuffering. The next encoded report clears that flag.

   The default play-state event therefore consumes bs immediately. Later response or interval reports during the same stall omit it. The first report after recovery can omit it too.

   Track whether starvation overlapped the interval since that destination's previous report. Include ongoing starvation when assembling bs, and preserve the final interval across recovery.

   The [bs definition](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#buffer-starvation) covers remaining in rebuffering, as well as entering it.

5. **[P2] Define startup tracking across rotation**

   [rfc/cmcd-session-api.md:361](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L361)

   Rotation resets the startup timestamp but preserves the reporter's stored play state. Transition tracking compares against that stored state.

   Consider update({ sta: 's', ts: 1000 }), followed by rotate('manifest-sid'). Another update({ sta: 's' }) cannot restart measurement because the stored state is already s. The later transition to p has no startup timestamp, so msd is never derived.

   This affects the proposed manifest-supplied sid workflow directly. Define whether rotation preserves or rebases an active startup measurement. Resetting event deduplication alone does not reset transition tracking.

   The conflicting steps are [transition tracking](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L129) and [rotation](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L208).

6. **[P2] Give supplied bsd values consumable list semantics**

   [rfc/cmcd-session-api.md:378](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L378)

   CmcdMetric accepts a scalar or one value per object type. It cannot represent multiple completed stalls with the same cause, such as bsd=(300;v 700;v).

   The persistent override rule also stores update({ bsd: 300 }) as a session fact. Later reports can replay that duration. The cursor consumes automatic spans, but no equivalent consumption rule exists for supplied durations.

   Give bsd a duration-list type. Specify how supplied samples enter destination-specific pending data and become consumed after reporting.

   The [bsd definition](https://cta-wave.github.io/Resources/common-media-client-data--cta-5004-b.html#buffer-starvation-duration) requires separate completed durations and once-only reporting per mode and destination.

7. **[P2] Pin the request origin before invoking transforms**

   [plans/cmcd-session-api/architecture.md:188](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L188)

   Decoration selects the request target before running its transform. It captures origin.sidState from current only after the transform returns.

   A transform can rotate the session. Decoration then uses the old target's report state but associates the response origin with the new sid state. The request record and its response no longer share a stable origin.

   Capture the sid state, reporter identity, and request data before running transforms. Use that origin throughout decoration and response registration. Define equivalent pinning for event fan-out.

   Transform-triggered rotation already has [regression coverage in PR #422](https://github.com/streaming-video-technology-alliance/common-media-library/pull/422#discussion_r3930552802). This interaction therefore needs an explicit contract in the replacement design.

8. **[P2] Preserve a drain request while a send is active**

   [plans/cmcd-session-api/architecture.md:225](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L225)

   processQueue(true) returns when another send is active. The algorithm stores no pending drain request.

   Consider a target with batchSize: 10. While one batch is in flight, three additional lines queue. Disposal requests a drain, which returns immediately. The successful send then resumes ordinary processing, leaving the partial batch below its threshold.

   Persist the drain request until the queue empties. Continue draining after the active request settles, including dropped client-error batches. Apply the rule to flush(), rotate(), and dispose().

   The same rule must cover lines appended after an ended state was initially drained. Object retention alone does not schedule their delivery.

9. **[P2] Expose the promised response overrides**

   [rfc/cmcd-session-api.md:388](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L388)

   recordResponse() accepts CmcdPlaybackData, but that type excludes response keys. CmcdResponseInfo offers only status, headers, and resource timing.

   Callers therefore cannot supply ttfbb or smrt. Neither key has a derivation rule. Callers also cannot use the promised direct overrides for ttfb, ttlb, or CMSD values.

   Add a response-specific input type, or expose the missing fields through CmcdResponseInfo. Keep reporter-owned identity and sequence fields excluded.

   The contradiction appears in the [method signature](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L348), [supplied-value promises](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L416), and [response input](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L513).

10. **[P2] Include generated headers in the decorated-request type**

   [rfc/cmcd-session-api.md:496](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L496)

   The return type intersects R with only the cmcd member. Calling decorate({ url }) infers R without headers.

   The quick-start access req.headers therefore fails with TS2339. I confirmed this using the repository's TypeScript compiler and the proposed generic declarations.

   Define the output URL, headers, and cmcd explicitly. Preserve other request fields through Omit<R, 'url' | 'headers' | 'cmcd'>. Replacing these fields also avoids retaining incorrect literal types from the input.

11. **[P2] Remove stale CMCD headers before redecorating**

   [plans/cmcd-session-api/architecture.md:187](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L187)

   Header decoration copies existing headers and replaces only nonempty shards. It never removes a shard that becomes empty.

   Re-decorating a request after removing object metrics therefore preserves the previous CMCD-Object header. A consumed bs flag can similarly remain in CMCD-Status. The wire then contains values absent from cmcd.data.

   Remove existing CMCD shards case-insensitively before inserting the new shards. Preserve unrelated headers. Define how existing CMCD data is removed when transmission mode changes.

12. **[P2] Reject unavailable browser timing values**

   [plans/cmcd-session-api/architecture.md:197](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L197)

   The response algorithm subtracts startTime from responseStart without checking whether responseStart is available.

   A cross-origin resource can have startTime: 5000 and responseStart: 0 when timing access is restricted. The algorithm produces ttfb: -5000.

   Define validity checks and omit unavailable timing metrics. Also require response-body completion before deriving ttlb from the recording time. The quick-start currently records immediately after fetch resolves.

   The [Resource Timing specification](https://www.w3.org/TR/resource-timing/#sec-cross-origin-resources) documents the zero values used when timing access fails.

13. **[P2] Validate batch size against queue capacity**

   [rfc/cmcd-session-api.md:300](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/rfc/cmcd-session-api.md#L300)

   The configuration accepts batchSize: 1000 with the default maxQueueSize: 500. The queue cap prevents the send threshold from being reached.

   A live session with this configuration can discard old reports indefinitely without sending an automatic batch.

   Reject incompatible sizes or define a threshold bounded by queue capacity. Validate finite positive integers for both settings. Document the handling of zero capacity.

14. **[P2] Reclaim completed starvation spans**

   [plans/cmcd-session-api/architecture.md:72](https://github.com/streaming-video-technology-alliance/common-media-library/blob/7ec32e4cde7d609599de848aee8923e8d000214c/plans/cmcd-session-api/architecture.md#L72)

   The spans array only grows. Reporting advances destination cursors but never removes consumed spans.

   A long-lived session therefore retains every completed starvation duration. Request-only version 1 sessions retain durations even though those sessions cannot transmit bsd.

   Remove prefixes consumed by all eligible destinations. Avoid retaining samples for destinations that can never report bsd. Define a bound or loss policy when a transform suppresses bsd indefinitely.

   This finding concerns the proposed persistent state. It does not require changing the choice to retain request origins through object lifetime.

The historical comments materially changed this review. I did not reopen the resolved naming, derive options, or background-event exit format. I also treated the documented URL-only fallback and interval policy as deliberate choices.

The prior destination-identity decision and snapshot guarantees remain relevant because this proposal cites those designs as predecessors. Their removal needs an explicit replacement contract.

Validation: the quick-start type probe fails with TS2339 as described above. No package build or runtime tests were applicable to these documentation-only changes. The repository working tree remained unchanged.

littlespex:
**Review of 2026-09-10 14:41, dispositions.** Commit d83a44f1 applies them to the RFC and the design record. All fourteen findings were checked against the documents, the spec text, the dash.js source, and the earlier reviews they cite.

1. **Request state per destination.** Declined. CTA-5004-B defines targets as the configured Event Mode endpoints, in items 4, 5, 8, 12 to 14, and 16 of section 5. It names no request-mode target below the mode itself, so the `sn` rule gives request mode one sequence. The `bs` note on a new CDN warns about carry-over inside one request-mode stream and does not split it. Rationale now records that the PR #398 identity is not carried over, and Future possibilities lists an optional destination name on `decorate()`.
2. **Immutable data for late responses.** Accepted. `rotate()` copies each reporter's store onto the ended `sid` state, a late `rr` for an ended `sid` reads that copy, and `decorate()` copies the per-request data. There is no retention cap. An ended state is reachable only through request origins in the `WeakMap`, so its lifetime is bounded by the request objects the player keeps.
3. **Configuration changes.** Accepted. `session.configure()` replaces `version`, `transmissionMode`, `keys`, and `headerMap` for the current and later `sid` states, with no emission and no counter reset. The dash.js manifest flow is `configure()`, `rotate(sid)`, `update({ cid })`. Event targets stay fixed at creation, which matches dash.js, where they come from settings only.
4. **`bs` during a stall.** Accepted. The commit step clears the flag only when the store's `sta` is not `r`, so every report during the stall and the first one after recovery carry it. Rotation sets it on the new targets when the player is rebuffering.
5. **Startup tracking across rotation.** Accepted. `rotate()` carries `msdStart` while `msd` is not yet derived or supplied. A stall open at rotation is measured by the new `sid` from the rotation time.
6. **`bsd` semantics.** Accepted in substance, with a different type outcome. Spec item 14 allows one value per cause in these lists, so `bsd=(300;v 700;v)` is not valid and the type stays `CmcdMetric`. Each completed stall is one pending sample, a report carries the oldest unsent sample per cause, and each destination consumes it once, in order. A supplied `bsd` appends a sample and stops automatic detection. Values are never summed.
7. **Origin pinned before transforms.** Accepted. `decorate()` captures the origin first, and event fan-out iterates the `sid` state that was current when the emission began.
8. **Drain request during a send.** Accepted. A per-target `drainRequested` flag is set by `flush()`, `rotate()`, `dispose()`, and a late `rr`. It is honored when the send settles and cleared when the queue is empty.
9. **Response overrides.** Accepted. `CmcdResponseData` is the data type of `recordResponse()`: the playback data plus `ttfb`, `ttlb`, `ttfbb`, `smrt`, `cmsds`, `cmsdd`, `rc`, and `url`.
10. **Decorated request type.** Accepted. `Omit<R, 'url' | 'headers' | 'cmcd'>` plus explicit `url`, optional `headers`, and `cmcd`.
11. **Stale CMCD placement.** Accepted. `decorate()` removes the `CMCD` parameter and every `CMCD-` header, case-insensitively, before it writes the current placement.
12. **Timing validity.** Accepted. `ttfb` is omitted for an absent, zero, or early `responseStart`, and `ttlb` when no usable end time exists. The clock fallback measures the call, so the quick start now reads the body before `recordResponse()`.
13. **Batch size against queue capacity.** Accepted. Both sizes must be finite positive integers, `batchSize` at most `maxQueueSize`, and `interval` must be finite.
14. **Span reclamation.** Accepted. Consumed prefixes are dropped once every eligible destination has passed them, no samples are kept when no destination can report `bsd`, and each per-cause list is capped at 100.

PR #456, opened 2026-09-11
fix(cmcd): encode a version 1 nor value once when baseUrl is set

littlespex:
## Summary

- `toUrlSafe` in `CMCD_FORMATTER_MAP` relativizes an absolute `nor` URL with `urlToRelativePath`. That function builds a `new URL()` and returns a percent-encoded path. `encodeURIComponent` on that path turned every `%` into `%25`. With `version: 1` and a `baseUrl`, `https://cdn.example.com/next seg.mp4` was sent as `nor="next%2520seg.mp4"`.
- The formatter now uses the `encodePath` rule from `formatNor` on the session API branch. An existing `%XX` escape stays as it is. Every other unsafe character goes through `encodeURIComponent`. The value is now `nor="next%20seg.mp4"`.
- The regex carries the `u` flag. Without it, an astral character in a path splits into surrogates and `encodeURIComponent` throws "URI malformed".
- Version 2 output is unchanged. The formatter does not percent-encode a version 2 value.

## Packages Changed

## Test Plan

- [x] `npm run build -w libs/cmcd` and `npm test -w libs/cmcd`: 607 tests pass, 4 of them new (absolute `nor` with a space plus `baseUrl`, already-escaped relative path, escaped slash, escaped ampersand in a query string). The first two failed with `%2520` before the fix. The last two fail under plain `encodeURIComponent`, which produces `%252F` and `%2526`
- [x] Root `npm test` passes on all 18 packages
- [x] `encodeCmcd({ nor: '../testing/3.m4v' }, { version: 1 })` still gives `nor="..%2Ftesting%2F3.m4v"`

Notes for review:

- One relative-input case changes on the wire. A version 1 `nor` of `next%20seg.mp4` was sent as `next%2520seg.mp4` and is now sent unchanged. That is the intended consequence of keeping existing escapes. The changelog states it.
- CTA-5004 defines `nor` as a relative path that "MUST be URLEncoded", in headers as well as in the query string. The encoding turns the path into its URL form and encodes the delimiters, as the spec example `nor="..%2F300kbps%2Fsegment35.m4v"` shows. A receiver decodes once and has a path, so version 1 cannot tell an escaped separator from a delimiter. The review thread has the comparison with `main`.
- This branch and `feat/cmcd-session-api` both add to `## [Unreleased]` in the cmcd changelog. That branch also has its own `encodePath` in `formatNor.ts`. Whichever merges second has a small changelog conflict, and the two helpers can then share one module.


libs/cmcd/src/CMCD_FORMATTER_MAP.ts
```diff
@@ -18,6 +18,10 @@ const toRounded = (value: CmcdValue) => {
 	return roundValue(value)
 }
 
+const ESCAPE_OR_UNSAFE = /%[0-9A-Fa-f]{2}|[^A-Za-z0-9\-_.!~*'()]/gu
+
+const encodePath = (path: string): string => path.replace(ESCAPE_OR_UNSAFE, part => part.length === 3 ? part : encodeURIComponent(part))
```
littlespex, line 23: **[P2] Preserve URL escapes through CMCD encoding**

Preserving every `%XX` removes the CMCD encoding layer around existing URL escapes. With version 1, `dir%2Fsegment.mp4` now becomes `dir/segment.mp4` after the receiver decodes `nor`. Similarly, `segment.mp4?token=a%26b` becomes `segment.mp4?token=a&b`.

These changes can cause prefetching to request another object or alter authentication parameters. Reserved characters and their escapes are not equivalent ([RFC 3986, section 2.2](https://www.rfc-editor.org/rfc/rfc3986#section-2.2)).

Keep encoding `%` as `%25`, and add tests covering encoded separators.

littlespex (commented):
Found one regression in the version 1 `nor` formatter. The inline comment includes reproductions and the suggested correction.

Validation on commit `0156db6f3ac8cbb48b178c3aad9392849d0f767c`:

- CMCD build passed.
- CMCD tests: 605 passed, zero failures, and two TODOs.
- Project typecheck and lint passed.
- The public API report is unchanged.

libs/cmcd/src/CMCD_FORMATTER_MAP.ts
littlespex: Checked this against the CTA-5004 version 1 text rather than version 2.

Table 1 defines `nor` as "Relative path of the next object to be requested… This MUST be a path relative to the current request. This string MUST be URLEncoded [5]." Rule 10 says header payloads "MUST NOT be URLEncoded", yet the spec's own header example is `CMCD-Request:nor="..%2F300kbps%2Fsegment35.m4v"`. So the `nor` encoding is required even where nothing else is encoded, and it encodes the path's delimiters. That is the step that turns a path into its URL form, not a transport layer around an already-URL-form reference. The wording "the relative path, as defined by RFC 3986" is version 2.

Under the version 1 rule a receiver decodes once and has a path. It cannot tell an escaped slash from a delimiter, because the format encodes delimiters too. `main` only preserved that difference by encoding an existing escape a second time. The spec does not describe that second layer, and it breaks the common input. A URL-form path from a manifest such as `next%20seg.mp4` went on the wire as `%2520`.

| Input | `main` after one decode | this branch after one decode |
|---|---|---|
| `next%20seg.mp4` from a manifest | `next%20seg.mp4`, wrong for a receiver that encodes the path | `next seg.mp4`, the path |
| `dir%2Fsegment.mp4` | `dir%2Fsegment.mp4`, only right if the receiver treats it as URL form | `dir/segment.mp4`, the escape is lost |

Neither output serves both receiver behaviors. The version 1 text picks the first row, and `formatNor` on the session API branch implements the same rule, so this branch keeps it. The escaped-separator tests are in 76126582 with the version 1 expectations: `dir%2Fsegment.mp4` stays `dir%2Fsegment.mp4`, and `segment.mp4?token=a%26b` becomes `segment.mp4%3Ftoken%3Da%26b`. The changelog entry now cites the spec rule.

littlespex:
Closing this PR because the existing formatter already preserves the intended URL in the reported case.

The original example works with the existing implementation:

1. The URL path is `next%20seg.mp4`.
2. Encoding the CMCD `nor` value produces `next%2520seg.mp4`.
3. Decoding `nor` restores `next%20seg.mp4`.
4. Resolving that path requests the intended object.

These are separate encoding operations. The `%2520` output does not demonstrate a defect.

I checked the supplied CTA-5004 version 1 specification and the response in the review discussion:

- Table 1, printed page 9, requires `nor` to be URLEncoded according to reference [5].
- Reference [5], printed page 18, identifies WHATWG form encoding. That encoding encodes `%` as `%25`.
- The header example encodes path delimiters but contains no existing escapes. It does not establish how existing escapes should be handled.
- The server requirements do not prescribe decoding into raw path characters and then reconstructing URL escaping.

The proposed formatter loses information. With the same host and `baseUrl`, these different absolute URL paths produce identical output:

| Input URL path | PR output for `nor` |
|---|---|
| `/dir/segment.mp4` | `dir%2Fsegment.mp4` |
| `/dir%2Fsegment.mp4` | `dir%2Fsegment.mp4` |

The existing formatter preserves the distinction by emitting `dir%252Fsegment.mp4` for the second path. The new tests assert the information loss instead of detecting it.

The Unicode change provides no independent fix. Existing `encodeURIComponent` already handles valid astral characters. The `u` flag addresses a risk introduced by the replacement regex.

Keep the existing formatter. Useful follow-up work would add tests that verify URL preservation and documentation explaining the encoding operations.

References: [WHATWG form encoding](https://url.spec.whatwg.org/#application/x-www-form-urlencoded-percent-encode-set), [RFC 3986 section 2.2](https://www.rfc-editor.org/rfc/rfc3986#section-2.2).

PR #461, opened 2026-09-12
fix(utils): remove every shared leading segment in urlToRelativePath

littlespex:
## Description

`urlToRelativePath` in `@svta/cml-utils` skipped shared leading path segments. The common-parent loop advanced its index while it also shifted both path arrays. It compared the wrong segments and removed segments it never compared.

A base directory with two segments, such as `/v/1080p/`, returned `../1080p/seg-2.m4s` for a target in the same directory. The correct result is `seg-2.m4s`. When the target and base diverged and matched again at a deeper segment, the old result resolved to a wrong URL. The `/v/1080p/` shape is common on CDNs. The bug reached CMCD `nor` values through `CMCD_FORMATTER_MAP`.

Reproduce on `main`:

```js
import { getBaseUrl, urlToRelativePath } from '@svta/cml-utils'

const base = getBaseUrl('https://cdn.example.com/v/1080p/seg-1.m4s')
urlToRelativePath('https://cdn.example.com/v/1080p/seg-2.m4s', base)
// main: "../1080p/seg-2.m4s"    this PR: "seg-2.m4s"
```

### Changes

- `urlToRelativePath` counts the shared prefix with a fixed index and stops before the destination file. The relative path is one `../` per remaining base directory followed by the remaining target segments. The new loop does no `shift` or `unshift` work.
- `urlToRelativePath` adds a `./` prefix when the result is empty, starts with `/`, or has `:` in its first segment (RFC 3986 section 4.2). Without the prefix, the reference resolved to a wrong URL. `main` had this bug for a one-segment or root base directory.
- Thirteen new tests in `libs/utils/test/urlToRelativePath.test.ts` cover the fix. The cases are one, two, and three common segments, a sibling directory, a subdirectory, a cross-origin target, and query and hash preservation. One more test covers a path that diverges and matches again at a deeper segment. Five tests cover the `./` prefix cases and resolve each result against the request URL with `new URL`.
- Changelog notes under Unreleased in `libs/utils` and `libs/cmcd`.

### Reviewer notes

**Wire output change.** CMCD `nor` values change for adopters whose `baseUrl` shares two or more leading segments with the request URL. Some values gain a `./` prefix. The cmcd changelog names both changes. The fix applies to `encodeCmcd`, `toCmcdHeaders`, `toCmcdQuery`, and `CmcdReporter` when `baseUrl` is set.

No existing CMCD test expectation changed. The only `baseUrl` tests share one common segment, and the old loop handled that case correctly.

`formatNor.ts` on the session API branch (`feat/cmcd-session-api`, PR #460) calls `urlToRelativePath` from `@svta/cml-utils`. It has no inlined copy of the algorithm. That branch inherits this fix through the dependency and needs no change.

The public API report for utils is unchanged.

### Verification

The tests were written first. Eight of the thirteen new tests failed against the old code for the predicted reasons before each fix. Root `npm test` passes: lint, build, typecheck, and every package test suite with 0 failures.

## Requirements Checklist

- [x] Docs/guides updated (no change needed: the TSDoc example and the API report are unchanged)
- [x] Changelog updated


libs/cmcd/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to
...
 - `validateCmcdEvents` and `validateCmcdEventReport` report an error when the body ends with a line feed. Every earlier `CmcdReporter` sent that line feed. A receiver that upgrades marks reports from those players as invalid
+- `nor` is the shortest relative path when the request URL and `baseUrl` share two or more leading path segments. `urlToRelativePath` in `@svta/cml-utils` skipped shared segments. A `baseUrl` directory with two segments, such as `/v/1080p/`, produced `nor=("../1080p/seg-2.m4s")` for a segment in the same directory. The corrected value is `nor=("seg-2.m4s")`. This is a wire output change. A receiver that compares `nor` strings sees the new value. When the request URL and `baseUrl` diverged and matched again at a deeper segment, the old value resolved to a wrong URL. The fix applies to `encodeCmcd`, `toCmcdHeaders`, `toCmcdQuery`, and `CmcdReporter` when `baseUrl` is set
```
littlespex: This changelog entry is too long. Condense it.

libs/utils/src/urlToRelativePath.ts
```diff
@@ -30,25 +30,15 @@ export function urlToRelativePath(url: string, base: string): string {
...
 	}
 
 	// add back paths
-	while (fromPath.length) {
-		fromPath.shift()
-		toPath.unshift('..')
-	}
-
-	const relativePath = toPath.join('/')
+	const relativePath = '../'.repeat(fromPath.length - common) + toPath.slice(common).join('/')
```
littlespex: **[P2] Preserve relative URL syntax after shortening**

With base `https://cdn.example.com/v/1080p/seg-1.m4s`, these valid targets now resolve incorrectly:

| Target pathname | New output | Problem |
|---|---|---|
| `/v/1080p/seg:2.m4s` | `seg:2.m4s` | Interpreted as a URL scheme |
| `/v/1080p//seg-2.m4s` | `/seg-2.m4s` | Resolves at the origin root |
| `/v/1080p/` | `""` | Resolves to the base file |

All three resolved correctly before this PR. CMCD also emits the incorrect values.

Prefix `./` when the result is empty, starts with `/`, or contains `:` in its first segment. Add tests that resolve the returned reference against the base with `new URL(result, base)`. The colon restriction follows [RFC 3986 §4.2](https://www.rfc-editor.org/rfc/rfc3986#section-4.2).

littlespex (commented):
Found one correctness regression worth fixing before merge. See the inline comment.

Verification: utils and CMCD builds passed. All 93 utils tests passed. CMCD tests reported 603 passes and two TODOs. Project typecheck passed.

libs/cmcd/CHANGELOG.md
littlespex: Condensed to 65 words in 4a961628. The root cause and the diverge-and-rematch case stay in the utils changelog. The entry gained one sentence for the `./` prefix from the other thread, because that prefix is also wire output.

libs/utils/src/urlToRelativePath.ts
littlespex: Fixed in 4a961628. The result gets a `./` prefix when zero `../` remain and it is empty, starts with `/`, or has `:` in its first segment (RFC 3986 section 4.2). Five new tests cover the colon, double-slash, directory-itself, and root-with-query cases plus a no-prefix boundary. Each test resolves the result with `new URL(result, requestUrl)` against the full request URL, not the directory. `new URL('', directory)` resolves to the directory and would hide the empty case. A CMCD receiver resolves `nor` against the request URL.

One correction to the table. The three targets resolved correctly on `main` only for a two-segment base. With a one-segment or root base, `main` emitted the same `seg:2.m4s`, `/seg-2.m4s`, and `""`, plus `?x=1` for a root target with a query. The prefix now covers every base depth. The utils changelog says so.

PR #464, opened 2026-09-15
fix(utils): encode base64 in 32 KB chunks

littlespex:
## Description

`encodeBase64` converts the bytes in 32 KB `subarray` views with `String.fromCharCode.apply`, then calls `btoa` once. The output does not change.

The function spread every byte into one `String.fromCharCode` call. The spread pushes each byte as an argument, so inputs above about 125 KB threw `RangeError: Maximum call stack size exceeded`. Small inputs paid for the iterator protocol.

| Input | main | this PR |
| --- | ---: | ---: |
| 32 bytes | 882 ns | 214 ns |
| 4 KB | 92 µs | 13 µs |
| 200 KB | `RangeError` | 266668 chars |

Review notes:

- `libs/utils/test/encodeBase64.test.ts` is new. Its 200 KB case failed before the fix. The TSDoc of `encodeBase64` now includes the example region of that file.
- `@svta/cml-structured-field-values` is the only package that calls `encodeBase64`. Its tests pass against the new build.
- `Uint8Array.prototype.toBase64` is not available in Node 24, so this PR does not use it.
- No version bump. The changelog note is under Unreleased.

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated


littlespex:
No actionable findings in commit `69da762`. Ready to merge.

Verified:

- Both affected packages build successfully.
- All 1,798 package tests pass.
- Project typecheck and changed-file lint pass.
- Chunk boundaries, offset views, and inputs through 10 MB match Node’s base64 output.
- Public API reports remain unchanged.

Validation used Node 24. Browser runtimes were not tested.

Issue #462, opened 2026-09-12
structured-field-values: encoder correctness gaps found by the performance audit


littlespex:
A performance audit of the `@svta/cml-structured-field-values` encoder found four correctness gaps. None of them is a performance issue. Each one needs a decision, because a fix changes behavior for callers that pass invalid input today. The audit is [plans/sfv-encoder-perf/audit.md](https://github.com/streaming-video-technology-alliance/common-media-library/blob/perf/sfv-encoder-allocations/plans/sfv-encoder-perf/audit.md) on the `perf/sfv-encoder-allocations` branch.

## 1. `serializeString` accepts non-ASCII characters

```js
serializeString('héllo') // returns "héllo"
```

RFC 8941 section 4.1.6 step 1 requires serialization to fail when the string cannot be converted to ASCII. The check today only rejects `%x00-1f` and `%x7f`. The `structured-field-tests` corpus has no non-ASCII vector, so the test suite cannot catch this gap.

## 2. `serializeDate` emits a fraction

```js
serializeDate(new Date(1500)) // returns @1.5
```

RFC 9651 section 4.1.10 serializes a Date as an Integer. The function divides `getTime()` by 1000 and passes the result to `serializeInteger`, which does not check for a whole number. Options: fail, or truncate toward zero.

## 3. `encodeSfDict` and `encodeSfList` reject a plain `SfInnerList` object

```js
encodeSfDict({ a: { value: [1, 2], params: {} } })
// throws: failed to serialize "{"value":[1,2],"params":{}}" as Bare Item
```

The `SfMember` type includes `SfInnerList`, so this call compiles. The serializer treats the object as a bare item and fails. Either the type should drop `SfInnerList`, or the serializer should accept the object form.

## 4. `sf-serialization.test.ts` cannot detect a missing failure

The test wraps encode and the `suite.canonical[0]` read in one `try`. If a `must_fail` vector encodes without an error, `canonical` is `undefined`, the read throws a `TypeError`, and the `catch` asserts `must_fail === true`, which passes. The encoder does fail on all 539 `must_fail` serialization vectors today, verified per vector. A future regression in that area would not be caught.

PR #465, opened 2026-09-15
perf(structured-field-values): encode without wrapper objects and repeated passes

littlespex:
## Description

The encoder of `@svta/cml-structured-field-values` allocates less and runs faster. The output does not change.

- `encodeSfDict`, `encodeSfList`, and `encodeSfItem` no longer wrap bare values in `SfItem`. `serializeDict` collects one part per member and joins once. `serializeList` and `serializeInnerList` append their members.
- `serializeString` returns a string that needs no escape after one regex test.
- `serializeKey` checks the character codes in a loop. `serializeToken` keeps its regex as a module constant without capture groups.
- `serializeParams` loops over `Object.keys` instead of entries, map, and join.
- `serializeInnerList` also accepts `(list, params)`. The `SfInnerList` object form still works. This overload is the only change in the API report.
- `serializeDecimal` compares the magnitude as a number. `serializeDecimal(1e21)` now fails instead of returning `1e+21.0`.
- New benchmark: `npm run bench -w libs/structured-field-values`. The header of `bench/bench.ts` lists the options.

`encodeSfDict` is about half of the cost of `encodeCmcd` for a 17-key request. The audit in [plans/sfv-encoder-perf/audit.md](https://github.com/streaming-video-technology-alliance/common-media-library/blob/perf/sfv-encoder-allocations/plans/sfv-encoder-perf/audit.md) has the analysis. The table shows `npm run bench -- --baseline=temp/dist/index.js`, where `temp/dist` is a copy of the `main` build. Median per call, then heap bytes allocated per call.

| Input | main | this PR | delta | bytes main | bytes this PR |
| --- | ---: | ---: | ---: | ---: | ---: |
| CMCD dict, 17 keys, bare values | 2949 ns | 1750 ns | -41% | 6530 | 2130 |
| CMCD dict, 17 keys, `SfItem` values | 2830 ns | 1899 ns | -33% | 5586 | 2130 |
| CMCD dict, 17 keys, `Map` | 2946 ns | 1695 ns | -43% | 7203 | 2003 |
| Dict, 100 integer keys | 10148 ns | 9016 ns | -11% | 26598 | 9833 |
| List, 20 integers | 649 ns | 544 ns | -16% | 1418 | 1346 |
| List, 10 URL strings | 2627 ns | 1887 ns | -28% | 3458 | 2197 |
| List, 10 strings with quotes and backslashes | 4319 ns | 4359 ns | +1% | 9613 | 8358 |
| Item with 3 parameters | 630 ns | 283 ns | -55% | 1514 | 418 |

Review notes:

- `npm run bench -- --baseline=temp/dist/index.js --verify` compared both builds on the bench inputs and on the whole `structured-field-tests` corpus: 1258 identical outputs and error messages. The decimal fix is not in the corpus and has its own test.
- One error text changes: `encodeSfItem([1, 2])` reports `[1,2]` instead of the wrapped items.
- The `Map` path uses `forEach`. It allocates about 1 KB less per 17-key dict than `for...of` and runs about 170 ns slower.
- Strings that need an escape pay one extra regex test, the +1% row.
- `serializeMember` and `serializeDictMember` are package private.
- A bare `import` of the package still bundles to only the `@svta/cml-utils` import line with Rollup and tsdown.
- No version bump. The changelog notes are under Unreleased.
- The audit found four correctness gaps that change behavior when fixed. They are in #462, not in this PR.

## Requirements Checklist

- [x] Docs/guides updated
- [x] Changelog updated

### DIFF
--- a/libs/structured-field-values/CHANGELOG.md
+++ b/libs/structured-field-values/CHANGELOG.md
@@ -12,10 +12,14 @@ and this project adheres to
 
 - README: the usage example prints its result instead of calling an undefined `assert`.
 - `parseIntegerOrDecimal` creates its `Error` only when parsing fails. Before this change, every call created an `Error` and captured a stack trace, also when parsing succeeded. Every Integer and Decimal in a parsed field passes through this function. Error messages do not change.
+- The encoder allocates less and runs faster. `encodeSfDict`, `encodeSfList`, and `encodeSfItem` no longer wrap bare values in `SfItem` objects, and the dictionary and parameter serializers no longer copy their entries through `Object.entries`, `Array.from`, and `map`. `serializeString` returns a string that needs no escape after one regex test. `serializeKey` checks the character codes in a loop. For a 17-key CMCD request dictionary, one encode takes about 40% less time and allocates 2.1 KB instead of 6.5 KB. The output does not change. The error for `encodeSfItem([1, 2])` now quotes the array as passed, `[1,2]`, instead of the wrapped items.
+- `serializeInnerList` also accepts the list and its parameters as two arguments: `serializeInnerList([1, 2], { a: 1 })`. The `SfInnerList` object form still works.
+- A benchmark for the encoder: `npm run bench -w libs/structured-field-values`. The header of `bench/bench.ts` describes the options.
 
 ### Fixed
 
 - Remove the module-scope template literal that built the `Integer or Decimal` error type name. Rolldown-based bundlers such as tsdown kept that statement and two constants in bundles that never used the parser. `parseIntegerOrDecimal` now builds the name inside the function. Error messages do not change.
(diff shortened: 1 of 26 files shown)


littlespex (commented):
Reviewed commit `2879ddba` against base `2b575de9`. I found no actionable issues introduced by this PR.

The encoder changes preserve supported serialization behavior. The new helpers remain package private, and the API report contains only the documented overload.

Validation on Node 24.16.0:

- Full repository build, lint, and typecheck passed.
- Structured-field-values, CMCD, and CMSD tests: 2,324 passed, zero failures, and two existing TODO tests.
- Benchmark verification against a fresh base build: all 1,258 cases produced identical outputs and error messages.
- A CMCD benchmark run confirmed lower encoding time and allocation for bare values, `SfItem` values, and `Map` inputs.

The existing correctness gaps documented in the audit remain outside this review.

